PR #8, opened 2019-07-23
Cookiecutter scaffolding

### LINKED ISSUE #4: Add project scaffolding
Add the project scaffolding from DAI-Lab cookiecutter

csala:
Resolve #4 

Major refactoring of the code to match the DAI-Lab cookiecutter structure and improve the benchmarking and Synthesizers User Interface.


README.md
```diff
@@ -1,264 +1,205 @@
...
-    </figure>
-    </td>
-    </tr>
-    </table>
-
-## Quickstart
-
-After installing the requirements and preparing the datasets, you only need to run the following
-commands to evaluate a synthesizer:
+The output of the `benchmark` function will be a `pd.DataFrame` containing all the scores
```
leix28: Should have a QuickStart to help people quickly run existing synthesizers on the Benchmark. Also, move QuickStart before installation for development.

sdgym/evaluate.py
```diff
@@ -0,0 +1,369 @@
...
+from sklearn.neural_network import MLPClassifier, MLPRegressor
+from sklearn.preprocessing import OneHotEncoder
+from sklearn.tree import DecisionTreeClassifier
+
+from sdgym.constants import CATEGORICAL, CONTINUOUS, ORDINAL
+
+LOGGER = logging.getLogger(__name__)
+
+
+_MODELS = {
```
leix28: In master branch, classifiers / regressors are tuned for each dataset. Please use the same settings as master branch.

setup.py
```diff
@@ -0,0 +1,112 @@
...
+from setuptools import find_packages, setup
+
+with open('README.md') as readme_file:
+    readme = readme_file.read()
+
+with open('HISTORY.md') as history_file:
+    history = history_file.read()
+
+install_requires = [
+    'certifi==2019.3.9',
```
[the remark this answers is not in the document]
csala: Done. This had been inherited from the old `requirements.txt` and was not reviewed.
I removed several libraries that were not used and specified safe ranges instead of fixed versions.

README.md
csala: I added some explanation about the repo synthesizers and added some instructions about how to use them in the Usage section.
Regarding naming: the section is called Usage instead of Quickstart because there isn't much else in the repository right now, so it's actually both a quickstart and quickend :-)
Regarding order, this is the standard order that we are using in all the other projects: overview -> install (pip + make install + make install-develop) -> data format -> quickstart (here, usage), so I prefer to keep it in this order.

sdgym/evaluate.py
csala: This is implicitly included in the `project_type` that we added to the metadata jsons.

README.md
```diff
@@ -25,6 +25,71 @@ Laboratory](https://dai.lids.mit.edu/) at MIT.
...
+    real_data: numpy.ndarray,
+    categorical_columns: list,
+    ordinal_columns: list
+) -> syntehtesized_data: numpy.ndarray
+```
+
+If your synthesizer implements a different interface, you can wrap it in a function like this:
+
+```python
+def my_synthesizer_function(real_data):
```
leix28: Add `categorical_columns`, `ordinal_columns`.

README.md
```diff
@@ -25,6 +25,71 @@ Laboratory](https://dai.lids.mit.edu/) at MIT.
...
+
+* `fit`: Fits the synthesizer on the data. Expects the following arguments:
+    * `data (numpy.ndarray)`: 2 dimensional Numpy matrix with the real data to learn from.
+    * `categorical_columns (list or tuple)`: List of indexes of the columns that are categorical within the dataset.
+    * `ordinal_columns (list or tuple)`: List of indexes of the columns that are ordinal within the dataset.
+* `sample`: Generates new data resembling the original dataset. Expects the following arguments:
+    * `n_samples (int)`: Number of samples to generate.
+* `fit_sample`: Fits the synthesizer on the dataset and then samples as many rows as there were in
+  the original dataset.
+
```
leix28: Maybe move line 49 here. Give a solid example on how to wrap existing synthesizer class to a synthesizer function. I would prefer using Synthesizer class instead of synthesizer function for two reasons: 
- Usually, a synthesizer requires 2 phases, fit and sample.
- It's possible to make a consistent save/restore API.

README.md
```diff
@@ -102,55 +167,30 @@ make install-develop
 Make sure to use them regularly while developing by running the commands `make lint` and
 `make test`.
 
-## What is a synthesizer function?
+## Compile C++ dependencies
```
leix28: We should check PrivBayes' license.

csala:
> I think one additional feature to have is to log the performance of epoch based synthesizers in every K epochs. Not sure if that is easy to implement.

The problem that we have is that this is something very specific of this synthesizers in particular.

I prefer keeping the `benchmark` function to be as simple and generic as possible, so all kinds of synthesizers can be evaluated using it, and just focus on one result: the last one.

In any case, intermediate results can also be evaluated by simply running the benchmark on a Synthesizer instance created with a lower number of epochs in the arguments.

README.md
csala: > Maybe move line 49 here.

No, this function needs to stay there, right where the synthesizer function is explained, as that  is the place where the clarification about how to use a different API needs to be. Otherwise, that explanation would be really confusing.

> Give a solid example on how to wrap existing synthesizer class to a synthesizer function.

The complete example about how to create an instance of a synthesizer and evaluate it using the benchmark function is further down, on the usage section, and I'd also prefer to keep it there, as adding some code examples before the actual basic benchmark usage example would also be very confusing.

However, what I did is change a bit the Usage section to also show, apart from the benchmark, how the `fit` and `sample` methods would be used on their own without running the benchmarking.

> I would prefer using Synthesizer class instead of synthesizer function for two reasons:
> 
> * Usually, a synthesizer requires 2 phases, fit and sample.
> * It's possible to make a consistent save/restore API.

This is already like that: the synthesizers included are all classes and they all have the `fit` and `sample` methods. They just provide the `fit_sample` method for convenience, to have something with the necessary signature to be able to be passed directly as an argument to the `benchmark` function without having to do any wrapping.

But this is slightly different than the `benchmark` function. This has been created as simple as possible to support as many different APIs as possible, and that's why it has been decided to make it expect a function as input: it's much easier to wrap a Class based synthesizer inside a function than to make any synthesizer that is not a class work as a class and support the fit + sample interface.

README.md
csala: Can you please do that? All I did was adding the compile target to the Makefile, but the privbayes folder was already there since the beginning, so I do not know where you got it from.

leix28:
Look good to me.

Issue #10, opened 2019-08-07
sdgym.benchmark function bug


johnson18404:
Hello DAI-Lab, recently I am researching on data synthesizing and I found this amazing project.
I got an error when running `sdgym.benchmark()` function. 

![image](https://user-images.githubusercontent.com/13941886/62620930-24a62780-b94d-11e9-82f7-1f6cd06c1afb.png)

After tracking some codes. I think the variable `synthesized` should be a synthesized data produced by `synthesizer.sample()`.
So we can solve this problem by replacing line 37:

```
synthesized = synthesizer(train, categoricals, ordinals)
```

into

```
synthesizer.fit(train, categoricals, ordinals)
synthesized = synthesizer.sample(train.shape[0])
```

csala:
Hi @johnson18404 Thanks for reporting this.

The problem that you see is that the `synthesizer` argument of the `benchmark` function is expected to be a `callable`, which means that it can either be a `function` or `method`, but not a class instance.

If you look at the [Benchmarking the Demo Synthesizers](https://github.com/DAI-Lab/SDGym#benchmarking-the-demo-synthesizers) section in the README, you will find an example about how to evaluate the included synthesizers using the `benchmark` function, which is by passing the `synthesizer.fit_sample` method to the `benchmark` function:

```
synthesizer = IndependentSynthesizer()
scores = benchmark(synthesizer.fit_sample)
```

johnson18404:
I get it. It seems that I misunderstood the argument meanings. Thank you.

PR #11, opened 2019-08-07
fix bug in sdgym.benchmark()

johnson18404:
Fix issue  https://github.com/DAI-Lab/SDGym/issues/10


csala:
Thanks for the interest and for the contribution @johnson18404 !

However, I'm afraid that #10 is not really an issue but rather a consequence of an unintended usage of the function, so I'm closing this PR for now and I'll put more details about the solution in the issue itself.

PR #35, opened 2020-04-09
New API and docs

csala:
Add a new API layer to obtain the benchmark results as a leaderboard.
Also reorganize the README and the docs.


BENCHMARK.md
```diff
@@ -0,0 +1,162 @@
...
+                        adult/accuracy  adult/f1  ...  ring/test_likelihood
+IndependentSynthesizer         0.56530  0.134593  ...             -1.958888
+UniformSynthesizer             0.39695  0.273753  ...             -2.519416
+IdentitySynthesizer            0.82440  0.659250  ...             -1.705487
+...                                ...       ...  ...                   ...
+my_synthesizer_function        0.64865  0.210103  ...             -1.964966
+```
+
+However, the `benchmark` function supports a number of additional arguments that allows us
+to configure its behavior to better adapt it to our needs.
```
leix28, line 28:
```
`benchmark` function supports a number of additional arguments. (See Doc)[https://sdv-dev.github.io/SDGym/api/sdgym.benchmark.html?highlight=benchmark#module-sdgym.benchmark]
```

BENCHMARK.md
```diff
@@ -0,0 +1,162 @@
...
+    Number of iterations to perform over each dataset and synthesizer. Defaults to 3.
+* `add_leaderboard (bool)`:
+    Whether to append the obtained scores to the previous leaderboard or not. Defaults
+    to `True`.
+* `leaderboard_path (str)`:
+    Path to where the leaderboard is stored. Defaults to the leaderboard included
+    with the package, which contains the scores obtained by the SDGym synthesizers.
+* `replace_existing (bool)`:
+    Whether to replace old scores or keep them in the returned leaderboard. Defaults
+    to `True`.
```
leix28, line 51: I would suggest to add an execution timestamp to the leaderboard, so that people can compare old and new results.

BENCHMARK.md
```diff
@@ -0,0 +1,162 @@
...
+* `datasets (list[str])`:
+    Names of the datasets to use for the benchmark. Defaults to all the ones available.
+* `iterations (int)`:
+    Number of iterations to perform over each dataset and synthesizer. Defaults to 3.
+* `add_leaderboard (bool)`:
+    Whether to append the obtained scores to the previous leaderboard or not. Defaults
+    to `True`.
+* `leaderboard_path (str)`:
+    Path to where the leaderboard is stored. Defaults to the leaderboard included
+    with the package, which contains the scores obtained by the SDGym synthesizers.
```
leix28, line 48: I would suggest not to overwrite the leaderboard inside the package. 

Basically, here are the expected behavior:

`add_leaderboard = False, leaderboard_path = None`: write new leader board to current directory.

`add_leaderboard = True, leaderboard_path = None`: write leader board to current directory, and append default leaderboard. 

`leaderboard_path is not None`: write leader board to a specified path. `add_leaderboard` decides overwriting or appending new results.

BENCHMARK.md
csala: It's already there :-)
So, if replace_existing is True, rows are duplicated with different timestamps.

BENCHMARK.md
csala: We are not overwriting the leaderboard. This is the path where the current (or reference) leaderboard is stored, but it is never modified.
The new leaderboard is always returned as a DataFrame, and the user can then store it where they need it using `to_csv` on it.

PR #50, opened 2020-09-24
Track execution times and report variation

### LINKED ISSUE #49: Track execution times and report variation
The leaderboard could be extended to keep track of execution times of the synthesizers and report the variation of the measurements (by default the three iterations). 

To some users, the execution time is an informative measure for evaluating the feasibility of using that synthesizer due to resource limitations (and also interesting in general). Reporting the variation for the measurements is necessary to be able compare synthesizers.

sbrugman-ing:
This PR resolves #49.

- Extends `benchmark` with `add_std` parameter, which if set to `True`, included the variation for the measurements.
- Keeps track of execution times of the synthesizers.


csala:
Hi @sbrugman-ing! Unfortunately, we have done some changes in the project structure that make this PR not compatible any more, so I'm closing it for now. Please, feel free to re-open it if you are able to adapt to the new project structure.

Also, I see that you are proposing several independent things on your PR. Would you mind opening separated issues/PRs for each one of them, so we can evaluate/discuss them separately?

PR #52, opened 2020-09-28
Reduce tox verbosity

### LINKED ISSUE #51: Travis CI tests failing
THe Travis CI tests are failing because of the following:

> The job exceeded the maximum log length, and has been terminated.

Example: [this job](https://travis-ci.org/github/sdv-dev/SDGym/jobs/730968391)

sbrugman:
Fixes #51

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,12 +16,12 @@ passenv = CI TRAVIS TRAVIS_*
 skipsdist = true
 skip_install = true
 commands_pre =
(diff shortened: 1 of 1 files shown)


csala:
Hi @sbrugman and thanks for the PR!

Unfortunately, after digging a bit into this topic we decided to move in the same direction as other libraries in the ecosystem and slightly changed the way tox creates the environment (#53 ), so this change will not be needed anymore.

Thanks again!

PR #74, opened 2021-05-10
Allow reading datasets from private s3 bucket

katxiao:
Add functionality to download datasets from a private s3 bucket.
- Add aws_key and aws_secret arguments to run function and use them on the download dataset. 
- `boto3` needs to be used instead of urllib.
- Also adapt unit tests as necessary if any of them fail.

✅ Verified that e2e flow works when pulling data from a public s3 bucket 
✅ Verified that e2e flow works when pulling data from a private s3 bucket

### DIFF
--- a/sdgym/datasets.py
+++ b/sdgym/datasets.py
@@ -1,11 +1,11 @@
 import io
 import logging
-import urllib.request
 from pathlib import Path
-from xml.etree import ElementTree
 from zipfile import ZipFile
 
 import appdirs
+import boto3
+import botocore
 import pandas as pd
 from sdv import Metadata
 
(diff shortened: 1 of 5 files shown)


sdgym/datasets.py
```diff
@@ -83,19 +94,36 @@ def load_tables(metadata):
...
-    datasets = []
-    for content in tree.findall('{*}Contents'):
-        key = content.find('{*}Key').text
-        size = int(content.find('{*}Size').text)
-        if key.endswith('.zip'):
-            datasets.append({
-                'name': key[:-len('.zip')],
-                'size': size
-            })
+    if S3_PREFIX in bucket_url:
```
csala, line 99: Same as above, the old `urllib` based approach can be removed, and `boto3` always used for the download and listing of datasets.

sdgym/datasets.py
```diff
@@ -17,21 +17,35 @@
...
+def _get_s3_client(aws_key=None, aws_secret=None):
+    if aws_key is not None and aws_secret is not None:
+        # credentials available
+        return boto3.client(
+            's3',
+            aws_access_key_id=aws_key,
+            aws_secret_access_key=aws_secret
+        )
+    else:
+        # no credentials available, make unsigned requests
```
csala, line 43: The current proposal has the problem that if credentials are available system-wide (or as environment variables) but `aws_key` or `aws_secret` are not passed, an anonymous client will be created, preventing access to a private bucket that would otherwise be available.

I would add a new `if` here, inside the `else`, which checks for system-wide credentials before falling back to the unsigned client.

```python3
else:
    if boto3.Session().get_credentials():
        # credentials available and will be detected automatically
        config = None
    else:
        # no credentials available, make unsigned requests
        config = botocore.config.Config(signature_version=botocore.UNSIGNED)

    return boto3.client('s3', config=config)
```

sdgym/datasets.py
```diff
@@ -17,21 +17,35 @@
...
+    obj = s3.get_object(Bucket=bucket, Key=f'{dataset_name}.zip')
+    bytes_io = io.BytesIO(obj['Body'].read())
 
     LOGGER.info('Extracting dataset into %s', datasets_path)
     with ZipFile(bytes_io) as zf:
         zf.extractall(datasets_path)
 
 
-def _get_dataset_path(dataset, datasets_path, bucket=None):
+def _get_s3_client(aws_key=None, aws_secret=None):
```
csala, line 34: This function should be placed above `download_dataset`

csala (changes requested):
This looks almost ready, but there is a situation that should be fixed to allow system-wide credentials to be used.

csala (changes requested):
Just one last code style change (line break after if/else block) and this is ready to go :-)

sdgym/datasets.py
```diff
@@ -17,21 +17,40 @@
...
+            aws_access_key_id=aws_key,
+            aws_secret_access_key=aws_secret
+        )
+    else:
+        if boto3.Session().get_credentials():
+            # credentials available and will be detected automatically
+            config = None
+        else:
+            # no credentials available, make unsigned requests
+            config = botocore.config.Config(signature_version=botocore.UNSIGNED)
```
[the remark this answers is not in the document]
katxiao: Done!

csala (approved):
Looks good!

PR #77, opened 2021-05-15
Store result files to s3

### LINKED ISSUE #81: Store cache contents into an S3 bucket
* SDGym version: 0.3.0

### Description

Similarly to what is described on #80, it should be possible to store the cache contents into an S3 bucket.

The behavior would be similar to the results path, where one can specify the `s3://` prefix in the `cache_dir` path specification to trigger the S3 storage.

Python

```python
sdgym.run(..., cache_dir='s3://my-bucket/path/to/my/cache/dir')
```

CLI

```bash
sdgym run ... -c s3://my-bucket/path/to/my/cache/dir
```

The `collect` command introduced in the PR #78 should also be adapted to read the cache contents from S3 and store the resulting CSV file to S3.

katxiao:
If the user passes `cache_dir` as an `s3://<bucket-name>/path/to/dir` the scores, synthetic data, and error files are stored in the corresponding path of the given bucket. If `aws_key` and `aws_secret` are provided, they will be used to authenticate the s3 requests.

✅ Ran with aws credentials and a private s3 bucket, and verified that the results files were uploaded.

Resolve #81

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -333,7 +333,8 @@ def run(synthesizers, datasets=None, datasets_path=None, modalities=None, bucket
         raise SDGymError("No valid Dataset/Synthesizer combination given")
 
     scores = pd.concat(scores)
+
     if output_path:
-        scores.to_csv(output_path, index=False)
+        write_csv(scores, output_path, aws_key, aws_secret)
 
     return scores
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -17,6 +17,7 @@ from sdgym.datasets import get_dataset_paths, load_dataset, load_tables
 from sdgym.errors import SDGymError
 from sdgym.metrics import get_metrics
 from sdgym.progress import TqdmLogger, progress
+from sdgym.s3 import is_s3_path, write_csv, write_file
 from sdgym.synthesizers.base import Baseline
 from sdgym.utils import (
     build_synthesizer, format_exception, get_synthesizers_dict, import_object, used_memory)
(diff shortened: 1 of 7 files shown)


sdgym/benchmark.py
```diff
@@ -149,11 +150,71 @@ def _score_with_timeout(timeout, synthesizer, metadata, metrics, iteration):
         return output
 
 
+def _write_to_cache_dir(cache_dir, name, dataset_name, iteration, run_id, scores,
```
csala, line 153: I'm thinking about changing this approach and, instead of writing a method that handles all the different situations, just write a `write_file(contents, path, aws_key, aws_secret)` function that knows whether to write in S3 or in a local file depending on the path prefix.
Additionally, a `write_csv` could be added that just dumps the DataFrame as a CSV string and then calls `write_file`.

With that, we can keep the code almost as it was before, doing:

```python3
base_path = str(cache_dir / f'{name}_{dataset_name}_{iteration}_{run_id}')
if scores is not None:
    write_csv(scores, base_path + '_scores.csv')
if 'synthetic_data' in output:
    synthetic_data = compress_pickle.dumps(output['synthetic_data'])
    write_file(synthetic_data, base_path + '.data.gz')
if 'exception' in output:
    write_file(output['exception'],base_path + '_error.txt')
```

sdgym/s3_utils.py
```diff
@@ -0,0 +1,48 @@
+import boto3
```
csala, line 1: I would rename this module to `s3.py`, as there is no other place where S3 is mentioned, so there is no ambiguity

csala (changes requested):

sdgym/s3.py
```diff
@@ -0,0 +1,146 @@
...
+        bucket_name, key = parse_s3_path(path)
+        s3.put_object(
+            Bucket=bucket_name,
+            Key=key,
+            Body=contents,
+            ContentEncoding=content_encoding,
+        )
+    else:
+        with open(path, write_mode) as f:
+            if write_mode == 'w':
```
katxiao, line 117: @csala I think you're right about writing text files with write mode `w` instead of `wb`. I updated it to special-case the `gzip` files to write mode `wb`.

csala (approved):
All looks good! Let's wait for the rest of the PRs to be merged to this, and then this can go in

PR #78, opened 2021-05-15
Add collect results command

### LINKED ISSUE #79: Add a way to collect cached results
* SDGym version: 0.3.0

### Description

Apart from producing a single dataframe or CSV file with the scores obtained by all the Synthesizers, SDGym has the option to store intermediate results, scores and error logs as the different tasks are run, which are kept inside the `cache_dir`. However, if the sdgym process is cut for some reason, there is no way to find all the intermediate results and put them together as a single CSV file again.

It would be interesting to have a `collect_results` function and an `sdgym collect` command that would do this job and allow producing a single scores CSV file from a collection of intermediate cached results.

katxiao:
Add collect.py to the sdgym folder and integrate it as an `sdgym collect` command.

Resolve #79

### DIFF
--- a/sdgym/__init__.py
+++ b/sdgym/__init__.py
@@ -12,6 +12,7 @@ __version__ = '0.3.1.dev2'
 
 from sdgym import benchmark, results, synthesizers
 from sdgym.benchmark import run
+from sdgym.collect import collect_results
 from sdgym.datasets import load_dataset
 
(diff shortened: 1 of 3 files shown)


csala (approved):
Thanks @katxiao this looks good! I created issue #79 to associate this PR to it.

Issue #80, opened 2021-05-17
Store SDGym results into an S3 bucket


csala:
* SDGym version: 0.3.0

### Description

Current SDGym implementation allows to produce a results table as either a DataFrame (when run from python) or as a CSV file stored in the local HDD.

It should also be possible to store the results in an S3 bucket, which would be triggered by passing an `output_path` that contains the `S3` prefix:

Python:

```python
sdgym.run(..., output_path='s3://my-bucket/path/to/my/results.csv')
```

CLI

```bash
sdgym run ... -o s3://my-bucket/path/to/my/results.csv
```

If the bucket is private, the AWS key and secret introduce in PR #74 should be used.

csala:
Closed via #83

PR #83, opened 2021-05-17
Allow output_path to be a s3 path

katxiao:
If the user passes `output_path` as an s3_path (`s3://<bucket-name>/path/to/dir`), the results are stored in the corresponding path of the given bucket. If `aws_key` and `aws_secret` are provided, they will be used to authenticate the s3 requests.

✅ Ran with aws credentials and a private s3 bucket, and verified that the output file was uploaded.

Resolve https://github.com/sdv-dev/SDGym/issues/80

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -333,7 +333,8 @@ def run(synthesizers, datasets=None, datasets_path=None, modalities=None, bucket
         raise SDGymError("No valid Dataset/Synthesizer combination given")
 
     scores = pd.concat(scores)
+
     if output_path:
-        scores.to_csv(output_path, index=False)
+        write_csv(scores, output_path, aws_key, aws_secret)
 
     return scores


sdgym/benchmark.py
```diff
@@ -392,7 +411,7 @@ def run(synthesizers, datasets=None, datasets_path=None, modalities=None, bucket
         raise SDGymError("No valid Dataset/Synthesizer combination given")
 
     scores = pd.concat(scores)
-    if output_path:
-        scores.to_csv(output_path, index=False)
+
+    _write_output_to_csv(scores, output_path, aws_key, aws_secret)
```
csala, line 415: After comments from #77 are addressed this can become a one line change to `write_csv(scores, output_path)`

csala (changes requested):

csala (approved):
Looks good!

PR #84, opened 2021-05-18
Allow collect to read from and write to s3

### LINKED ISSUE #85: Collect cached results from s3 bucket
* SDGym version: 0.3.0

### Description
In issue #79 we added a way to collect results from a collection of intermediate cached results to a single scores CSV file.

It should also be possible to read the intermediate cached results from an S3 bucket, and to store the results csv to a S3 bucket. 

Python:

```python
sdgym.collect.collect_results(input_path='s3://my-bucket/path/to/results', output_file='s3://my-bucket/path/to/my/results.csv')
```

CLI

```bash
sdgym collect ... -i s3://my-bucket/path/to/results -o s3://my-bucket/path/to/my/results.csv
```

If the bucket is private, the AWS key and secret introduce in PR #74 should be used.

katxiao:
Add the option to download the files from S3 and upload the results to S3. If `input_path` is a s3 path, we will read the results from s3 with the provided credentials. If `output_file` is a s3 path, we will upload the results file to s3 with the provided credentials.

✅ Tested with a private s3 bucket.

Resolves #85

### DIFF
--- a/sdgym/collect.py
+++ b/sdgym/collect.py
@@ -1,10 +1,7 @@
-import pathlib
-
-import pandas as pd
-import tqdm
+from sdgym.s3 import read_csv_from_path, write_csv
 
 
-def collect_results(input_path, output_file):
+def collect_results(input_path, output_file=None, aws_key=None, aws_secret=None):
     """Collect the results in the given input directory, and
     write all the results into one csv file.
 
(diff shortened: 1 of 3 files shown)


sdgym/collect.py
```diff
@@ -16,20 +19,47 @@ def collect_results(input_path, output_file):
...
 
     if output_file:
         output = output_file
     else:
-        output = run_path / 'results.csv'
+        output = f'{input_path}/results.csv'
 
     print(f'Storing results at {output}')
-    scores.to_csv(output, index=False)
+    if is_s3_path(output):
```
csala, line 55: After addressing the comments form PR #77, this can be simplified to just calling `write_csv(scores...)` instead of `scores.to_csv`

sdgym/collect.py
```diff
@@ -16,20 +19,47 @@ def collect_results(input_path, output_file):
...
+            access key will be used to read from and/or write to
+            any s3 paths.
     """
     print(f'Reading results from {input_path}')
     run_path = pathlib.Path(input_path)
 
     scores = []
-    for path in tqdm.tqdm(list(run_path.glob('**/*.csv'))):
-        scores.append(pd.read_csv(path))
+    if is_s3_path(input_path):
```
csala, line 35: Just like the `write_file` function, I would move this to the `s3` module as a `read_file` function (and maybe also `read_csv`) that is able to decide between reading from S3 and reading from a local path.

csala (changes requested):

csala (approved):
Good to go!

PR #86, opened 2021-05-18
Add unit tests for s3 functions

katxiao:
Add unit tests for the s3 module.

### DIFF
--- a/tests/unit/test_s3.py
+++ b/tests/unit/test_s3.py
@@ -1,0 +1,213 @@
+from unittest.mock import Mock, patch
+
+import pandas as pd
+
+from sdgym.s3 import is_s3_path, parse_s3_path, write_csv, write_file
+
+
+def test_is_s3_path_with_local_dir():
+    """Test the ``sdgym.s3.is_s3_path`` function with a local directory.
+
+    If the path is not an s3 path, it should return ``False``.
+
+    Input:
+    - path to a local directory
+
+    Output:
+    - False
+    """
+    # setup
+    path = 'path/to/local/dir'
+
+    # run
+    result = is_s3_path(path)
+
+    # asserts
+    assert not result
+
+
+def test_is_s3_path_with_s3_bucket():
+    """Test the ``sdgym.s3.is_s3_path`` function with an s3 directory.
+
+    If the path is an s3 path, it should return ``True``.
+
+    Input:
+    - path to an s3 directory
+
+    Output:
+    - True
+    """
+    # setup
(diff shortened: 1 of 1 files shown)


tests/unit/test_s3.py
```diff
@@ -0,0 +1,204 @@
...
+    - path
+
+    Output:
+    - boolean
+    """
+    # setup
+    path = 'path/to/local/dir'
+
+    # run
+    res = is_s3_path(path)
```
csala, line 22: very minor style comment here: We should try to use the entire variable names whenever possible. So let's call this `result` or `response`, rather than just `res`.

tests/unit/test_s3.py
```diff
@@ -0,0 +1,204 @@
+from unittest.mock import Mock, patch
+
+import pandas as pd
+
+from sdgym.s3 import is_s3_path, parse_s3_path, write_csv, write_file
+
+
+def test_is_s3_path_with_local_dir():
+    """Test the ``sdgym.s3.is_s3_path`` function with a local
```
csala, line 9: A small style detail: let's try to follow the docstring format recommendations from [PEP 257](https://www.python.org/dev/peps/pep-0257/), which means that the first line is a complete sentence, ending with a dot `'.'`, and any additional information is in a line below separated by a blank line.

Here I would say
```
"""Test the ``sdgym.s3.is_s3_path`` function with a local path.

If the given path is not an S3 path, it should return ``False``.
```

tests/unit/test_s3.py
```diff
@@ -0,0 +1,204 @@
...
+    directory. Expect that the return value is false.
+
+    Input:
+    - path
```
csala, line 13: I would try to be a bit more verbose about the input and output specification. More than just the type, we should say what is being tested (what is being passed) and how it is being validated (what do you expect to happen).

Here I would say:
```
Input:
- path to a local file
Output:
- False
```

tests/unit/test_s3.py
```diff
@@ -0,0 +1,204 @@
...
+
+    Output:
+    - bucket_name
+    - key_prefix
+    """
+    # setup
+    expected_bucket_name = 'my-bucket'
```
csala, line 61: I would define the `expected...` variables below, within the `assert` section of the test.

csala (changes requested):
The tests themselves look good and are testing the functions properly, but I added a few comments regarding code style and docstrings.

csala (approved):
Thanks @katxiao It looks good now!

PR #90, opened 2021-06-01
Add ydata to sdgym

fealho:
Adds [ydata](https://github.com/ydataai/ydata-synthetic) to our benchmark.


csala (approved):
Thanks @fealho This looks good to go!

PR #92, opened 2021-06-03
Add summary function and command

amontanez24:


sdgym/summary.py
```diff
@@ -3,6 +3,8 @@
 import numpy as np
 import pandas as pd
 
+from sdgym.results import add_sheet
```
csala, line 6: If the `results.py` module is not being used anywhere else (which I think is not), I would copy whatever is necessary here and remove the module.

csala (changes requested):
I added a few suggestions to improve the current implementation and interface

sdgym/summary.py
```diff
@@ -150,3 +153,54 @@ def errors_summary(data):
         all_errors[synthesizer] = errors.fillna(0).astype(int)
 
     return all_errors
+
+
+def make_summary_spreadsheet(file_path):
```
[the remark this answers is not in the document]
csala: @amontanez24 we should also add support to read from and write to S3 buckets, following an approach similar to the `collect` command.
This means we need to add `aws_key` and `aws_secret` arguments, and that we need to use the `s3` module functions to read and write files (we may need to change something there to write the xlsx file?)

sdgym/summary.py
```diff
@@ -198,9 +233,45 @@ def make_summary_spreadsheet(file_path):
...
+        baselines (dict):
+            Optional dict mapping modalities to a list of baseline
+            model names. If not provided, a default dict is used.
+    """
+    results = read_csv(results_csv_path, aws_key, aws_secret)
+    data = preprocess(results)
+    baselines = baselines or MODALITY_BASELINES
+    output_path = output_path or re.sub('.csv$', '.xlsx', results_csv_path)
+    output = io.BytesIO()
+    writer = pd.ExcelWriter(output) if aws_key and aws_secret else pd.ExcelWriter(output_path)
```
csala, line 269: `aws_key` and `aws_secret` cannot be used to decide whether we are writing in S3 or not, since an S3 path can be passed without keys to let `boto3` use the system-wide credentials.

To keep things simple, I would just not distinguish between a local or an S3 path here and always write to the `BytesIO` and pass it down to `write_file`, which will decide whether to write to a local or remote file based on the given path.

csala (changes requested):
Comment about how to decide whether to write to S3 or not, and how to do it.

sdgym/summary.py
amontanez24: I see, I didn't know that. This is nice though since it will be cleaner in the end

csala (approved):
Good to go! Thanks @amontanez24 !

PR #93, opened 2021-06-04
Allow jobs list/JSON to be passed

fealho:


sdgym/utils.py
```diff
@@ -74,29 +78,42 @@ def _get_synthesizer(synthesizer, name=None):
...
+                LOGGER.info('Trying to import synthesizer from fully qualified name.')
                 synthesizer = import_object(synthesizer)
+
             except Exception:
-                raise SDGymError(f'Unknown synthesizer {synthesizer}') from None
+                try:
+                    LOGGER.info('Trying to load synthesizer from JSON string.')
+                    return json.loads(synthesizer)
+
+                except JSONDecodeError as e:
```
csala, line 100: This try/except block is a bit strange, and I suspect that the `except Exception` part would never be reached, because at this point any string that is passed to the `json.loads` that is not a valid JSON will make it raise the `JSONDecodeError`.

Further than that, I would invert the order of things and do:
1. Try to load a JSON
2. If loading a JSON is not possible, try to import the synthesizer
3. If importing fails, it is an unknown synthesizer

tests/integration/test_benchmark.py
```diff
@@ -13,3 +13,23 @@ def test_identity():
     scores = output.groupby('synthesizer').score.mean().sort_values()
 
     assert ['Uniform', 'Independent', 'Identity'] == scores.index.tolist()
+
+
+def test_identity_jobs():
```
csala, line 18: Can we also add a test to validate the passing of a single synthesizer as a serialized JSON?

csala (changes requested):

csala (approved):
Good to go!

PR #95, opened 2021-06-09
Summary command fixes

katxiao:
Two small fixes to the summary command
- The args to `make_summary_command` weren't being passed in the expected order, so the `aws_key` was being passed in as the `baselines` argument.
- The `width` calculation was getting an error `*** TypeError: 'int' object is not iterable` when the dataframe was empty.

### DIFF
--- a/sdgym/__main__.py
+++ b/sdgym/__main__.py
@@ -137,8 +137,8 @@ def _collect(args):
 
 
 def _summary(args):
-    sdgym.summary.make_summary_spreadsheet(args.input_path, args.output_file, args.aws_key,
-                                           args.aws_secret)
+    sdgym.summary.make_summary_spreadsheet(args.input_path, output_path=args.output_file,
+                                           aws_key=args.aws_key, aws_secret=args.aws_secret)
 
 
 def _get_parser():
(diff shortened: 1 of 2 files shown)


csala (approved):
It looks good! Thanks @katxiao !

sdgym/__main__.py
```diff
@@ -137,8 +137,8 @@ def _collect(args):
 
 
 def _summary(args):
-    sdgym.summary.make_summary_spreadsheet(args.input_path, args.output_file, args.aws_key,
-                                           args.aws_secret)
+    sdgym.summary.make_summary_spreadsheet(args.input_path, output_path=args.output_file,
```
amontanez24, line 140: good catch!

sdgym/summary.py
```diff
@@ -173,7 +173,11 @@ def add_sheet(dfs, name, writer, cell_fmt, index_fmt, header_fmt):
 
         for idx, column in enumerate(df.columns):
             worksheet.write(startrow, idx, column, header_fmt)
-            width = max(len(column), *df[column].astype(str).str.len()) + 1
+            if df.empty:
```
amontanez24, line 176: Should we add a unit test for this?

amontanez24 (commented):
This looks good to me! Just a question on if one of us should add a unit test

sdgym/summary.py
csala: Good point @amontanez24
At some point we may add unit tests to cover all this code. But in this case, since the rest of the code is not tested either, I think it is fine to go ahead and merge as it is.

PR #96, opened 2021-06-10
Add max rows and max columns args

katxiao:
Add max rows and max columns args

### DIFF
--- a/sdgym/datasets.py
+++ b/sdgym/datasets.py
@@ -47,22 +47,29 @@ def _get_dataset_path(dataset, datasets_path, bucket=None, aws_key=None, aws_sec
     return dataset_path
 
 
+def _apply_max_columns_to_metadata(metadata, max_columns):
+    tables = metadata['tables']
+    for table in tables.values():
+        fields = table['fields']
+        if len(fields) > max_columns:
+            fields = dict(itertools.islice(fields.items(), max_columns))
+            table['fields'] = fields
+        structure = table.get('structure')
+        if structure:
+            structure['structure'] = structure['structure'][:max_columns]
+            structure['states'] = structure['states'][:max_columns]
+
+
 def load_dataset(dataset, datasets_path=None, bucket=None, aws_key=None, aws_secret=None,
                  max_columns=None):
     dataset_path = _get_dataset_path(dataset, datasets_path, bucket, aws_key, aws_secret)
-    metadata_file = open(str(dataset_path / 'metadata.json'))
-    metadata_content = json.load(metadata_file)
+    with open(dataset_path / 'metadata.json') as metadata_file:
+        metadata_content = json.load(metadata_file)
+
     if max_columns:
-        for table in metadata_content['tables']:
-            fields = metadata_content['tables'][table]['fields']
-            if len(fields) > max_columns:
-                fields = dict(itertools.islice(fields.items(), max_columns))
-                metadata_content['tables'][table]['fields'] = fields
-            if 'structure' in metadata_content['tables'][table]:
-                metadata_content['tables'][table]['structure']['structure'] = \
-                    metadata_content['tables'][table]['structure']['structure'][:max_columns]
-                metadata_content['tables'][table]['structure']['states'] = \
-                    metadata_content['tables'][table]['structure']['states'][:max_columns]
+        if len(metadata_content['tables']) > 1:
+            raise ValueError('max_columns is not supported for multi-table datasets')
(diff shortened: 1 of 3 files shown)


sdgym/datasets.py
```diff
@@ -45,9 +47,24 @@ def _get_dataset_path(dataset, datasets_path, bucket=None, aws_key=None, aws_sec
     return dataset_path
 
 
-def load_dataset(dataset, datasets_path=None, bucket=None, aws_key=None, aws_secret=None):
+def load_dataset(dataset, datasets_path=None, bucket=None, aws_key=None, aws_secret=None,
+                 max_columns=None):
     dataset_path = _get_dataset_path(dataset, datasets_path, bucket, aws_key, aws_secret)
-    metadata = Metadata(str(dataset_path / 'metadata.json'))
+    metadata_file = open(str(dataset_path / 'metadata.json'))
```
csala, line 53: This should be done inside a `with` block instead, so the file is closed after we are done reading it.
Also, I think the `str` call is not needed here.

sdgym/datasets.py
```diff
@@ -45,9 +47,24 @@ def _get_dataset_path(dataset, datasets_path, bucket=None, aws_key=None, aws_sec
...
+    metadata_content = json.load(metadata_file)
+    if max_columns:
```
csala, line 55: Can we move all of this inside a dedicated function?

Also, since relationships can be badly broken if we cut the number of columns, I would say that we should support this only for `single-table` and `time-series` datasets. Can we raise a `ValueError` if `max_columns` is defined and the dataset has more than one table?

sdgym/datasets.py
```diff
@@ -45,9 +47,24 @@ def _get_dataset_path(dataset, datasets_path, bucket=None, aws_key=None, aws_sec
...
+        for table in metadata_content['tables']:
```
csala, line 56: I would avoid 2 things here:
1. Accessing multiple times the same dict element (in this case `tables`). Instead, it is better to capture it in a variable.
2. Iterating over the dictionary keys to just get the corresponding values. Let's iterate over the `values` instead. (If the table name is needed at any point, then iterate over `items` capturing the name and table in a single shot.

```python3
tables = metadata_content['tables']
for table in tables.values():
    fields = table['fields']
    ...
```

sdgym/datasets.py
```diff
@@ -45,9 +47,24 @@ def _get_dataset_path(dataset, datasets_path, bucket=None, aws_key=None, aws_sec
...
+            fields = metadata_content['tables'][table]['fields']
+            if len(fields) > max_columns:
+                fields = dict(itertools.islice(fields.items(), max_columns))
+                metadata_content['tables'][table]['fields'] = fields
+            if 'structure' in metadata_content['tables'][table]:
+                metadata_content['tables'][table]['structure']['structure'] = \
```
csala, line 62: I would avoid 3 things here:
- Using `\` to create multi-line statements. If necessary, create an intermediate variable to hold the values temporarily.
- (Same as above): Accessing multiple times the same dict elements. Instead, let's capture the element in a variable (in this case, `structure`), and then use it as many times as needed. This will also help in the point mentioned above.
- Using `in` check on a dict element that you will end up getting from the dict if found. Instead, just directly take it using a `get` and then just check if that was `None` or not.

So, altogether, I would write this block of code like this:

```python3
structure = table.get('structure')
if structure:
    structure['structure'] = structure['structure'][:max_columns]
    structure['states'] = structure['states'][:max_columns]
```

sdgym/datasets.py
```diff
@@ -69,9 +86,11 @@ def load_dataset(dataset, datasets_path=None, bucket=None, aws_key=None, aws_sec
     return metadata
 
 
-def load_tables(metadata):
+def load_tables(metadata, max_rows=None):
     real_data = metadata.load_tables()
     for table_name, table in real_data.items():
+        if max_rows and table.shape[0] > max_rows:
```
csala, line 92: You can skip this check and directly get `table.head(max_rows)`, since `table.head(None) == table`, and `table.head(len(table) + K > 0) == table`

csala, line 92: Same as above about allowing `max_columns` only if this is a single table datasets: If there are more than one table and max_rows is defined, let's raise a ValueError saying that it is not supported for multi-table datasets.

csala (changes requested):
The approach is fine, but I made a few suggestions to clean up and improve the implementation, as well as a couple of validation checks

csala (approved):
Looks good @katxiao !
This can be merged as soon as the checks pass

PR #98, opened 2021-06-10
Automatically determine the number of workers

### LINKED ISSUE #97: Automatically detect number of workers
* SDGym version: 0.3.1

### Description

SDGym can automatically determine how many workers to use, based on the available GPUs or CPUs on the current machine. If the machine has GPUs, use the number of GPUs as the workers value. Otherwise use the number of CPUs. The user can request to automatically detect the number of workers by passing in `workers=-1`.

katxiao:
Automatically determine the number of workers based on the number of GPUs (if available) or the number of CPU cores.

Resolves #97

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -11,6 +11,7 @@ from pathlib import Path
 import compress_pickle
 import numpy as np
 import pandas as pd
+import torch
 import tqdm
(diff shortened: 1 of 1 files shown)


csala (approved):
It looks good @katxiao Merging this

PR #99, opened 2021-06-11
Updates rdt and sdv versions

fealho:


csala:
Not needed anymore after #106

PR #104, opened 2021-06-22
Initialize variable before reference

katxiao:
Fix bug `UnboundLocalError: local variable 'normalized_score' referenced before assignment` by initializing `normalized_score` to `None`.


csala (approved):
good catch

PR #106, opened 2021-06-28
Curate dependencies and make Gretel optional

csala:
Curate the dependencies and make Gretel optional to avoid conflicts.


katxiao (approved):
Nice!

PR #110, opened 2021-07-07
List synthesizer names

### LINKED ISSUE #82: List Synthesizer names
* SDGym version: 0.3.0
* Python version:
* Operating System:

### Description

Add a function and CLI command to list the available synthesizer names.

fealho:
Resolves #82.

### DIFF
--- a/sdgym/__main__.py
+++ b/sdgym/__main__.py
@@ -13,6 +13,8 @@ import tabulate
 import tqdm
 
 import sdgym
+from sdgym.synthesizers import __all__ as _synthesizers
+from sdgym.utils import get_synthesizers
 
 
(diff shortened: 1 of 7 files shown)


csala (changes requested):
@fealho I think we should roll back to using the `Baseline.get_subclasses`, or more than that, add a new `get_baselines` function which uses `Baseline.get_subclasses(include_parents=True)` and filters out the ones that are abstract.
For this, we will need to also add the `ABC` parent to the ones that are truly abstract, like the `Baseline` class, `SingleTableBaseline`...

csala (approved):
Thanks @fealho This looks good!

PR #119, opened 2021-08-20
Fix meta.yaml dependencies

fealho:
Corrects some dependency issues with the meta.yaml file. Specifically, changes `torch` to `pytorch`, because anaconda's version of `torch` is called `pytorch`, and removes `compress-pickle` because that package is not available through anaconda.


sarahmish (approved):
LGTM!

PR #126, opened 2021-11-04
Update dependency ranges for pandas and scipy

### LINKED ISSUE #118: Upgrade dependency ranges
The latest versions of the `compress-pickle` and `humanfriendly` are not supported by `SDGym`:

| Library | Upper bound (unsupported) | Latest release |
|-|-|-|
| `pandas` | 1.1.5 | 1.3.1 |
| `pomegranate` | 0.14.2 | 0.14.5 |
| `compress-pickle` | 2 | 2.01 |
| `humanfriendly` | 9 | 9.2 |

We should investigate why and update the code if necessary to support them.

katxiao:
Resolve #118 

I thought `pandas` was dependent on another one of our repos, but I think we can actually bump it directly. Same with `scipy`.


setup.py
```diff
@@ -18,12 +18,12 @@
     'compress-pickle>=1.2.0,<2',
     'humanfriendly>=8.2,<9',
     'numpy>=1.18.0,<2',
-    'pandas>=1.1,<1.1.5',
+    'pandas>=1.1,<2',
```
fealho, line 21: Huh, I'm surprised this works, I thought SDV had some issue with pandas 2.0, which is why we didn't upgrade it sooner.

Did you check that the other libraries work with pandas 2?

katxiao: I updated Copulas to use the latest version of pandas (1.3.4) here https://github.com/sdv-dev/Copulas/pull/279 and all the tests passed. I saw that you had mentioned some test case failures in Copulas but I couldn't reproduce it for some reason.

csala: I suspect that the problem in Copulas happened only in one specific version, which I think was `1.1.5`, because with later versions the problem seemed to disappear. But the strange thing is that now 1.1.5 does not seem to fail either when forced, so we can just forget about it and move on for now as @katxiao is suggesting.

fealho (approved):

csala (approved):

fealho (approved):
lgtm!

PR #132, opened 2021-12-27
Updates and bugfixes

csala:
Fix a bunch of errors that accumulated over time and also update a few dependencies.

List of changes are:
- Adapt LegacySingleTableBaseline to the latest RDT api by transforming only when categoricals are present in the data and also produce the right column names when sampling.
- Fix an argument value in GaussianCopulaOneHot which made it crash: `OneHotEncodingTransformer` -> `one_hot_encoding`
- Update to the latest gretel-synthetics version
- Update to the latest ydata-synthetic version, fix a few bugs (like using the wrong class name `VanillaGAN` -> `VanilllaGAN`) and add new models
- Adapt the Dockerfile to latest dependency versions and installation commants
- Fix logging problems created by the latest tensorflow and absl-py releases
- Default to running on all the synthesizers if none are given
- Fix error when writting an xlsx file to a local path

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,8 @@
 FROM nvidia/cuda:11.0.3-cudnn8-devel-ubuntu18.04
 CMD nvidia-smi
 
-RUN apt-get update && apt-get install -y build-essential && apt-get -y install curl
-RUN apt-get -y install python3.8 python3-distutils && ln -s /usr/bin/python3.8 /usr/bin/python
+RUN apt-get update && apt-get install -y build-essential curl python3.7 python3.7-dev \
+    python3-distutils && ln -s /usr/bin/python3.7 /usr/bin/python
 RUN curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py && \
     python get-pip.py && ln -s /usr/bin/pip3 /usr/bin/pip
 
@@ -18,8 +18,9 @@ COPY /privbayes/ /SDGym/privbayes
 WORKDIR /SDGym
 
 # Install project
-RUN make install-all compile
-RUN pip install -U numpy==1.20
+RUN pip install ydata-synthetic==0.6.1
+RUN pip install .[gretel] --no-binary pomegranate
+RUN make compile
 ENV PRIVBAYES_BIN /SDGym/privbayes/privBayes.bin
 ENV TF_CPP_MIN_LOG_LEVEL 2
 
--- a/Makefile
+++ b/Makefile
@@ -104,8 +104,13 @@ install-gretel-develop: clean-build clean-compile clean-pyc compile ## install t
 
 .PHONY: install-all
 install-all: clean-build clean-compile clean-pyc compile ## install the package with gretel and ydata
-	pip install 'ydata-synthetic>=0.3.0,<0.4'
+	pip install 'ydata-synthetic==0.6.1'
 	pip install .[gretel]
+
+.PHONY: install-all-develop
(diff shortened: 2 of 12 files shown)


Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM nvidia/cuda:11.0.3-cudnn8-devel-ubuntu18.04
 CMD nvidia-smi
 
-RUN apt-get update && apt-get install -y build-essential && apt-get -y install curl
-RUN apt-get -y install python3.8 python3-distutils && ln -s /usr/bin/python3.8 /usr/bin/python
+RUN apt-get update && apt-get install -y build-essential curl python3.7 python3.7-dev \
```
katxiao, line 4: Just curious, what's the reason for changing to python3.7?

sdgym/synthesizers/sdv.py
```diff
@@ -59,16 +59,16 @@ class GaussianCopulaOneHot(SDVTabular):
 
     _MODEL = sdv.tabular.GaussianCopula
     _MODEL_KWARGS = {
-        'categorical_transformer': 'OneHotEncodingTransformer'
+        'categorical_transformer': 'one_hot_encoding'
```
katxiao, line 62: Good catch!

Dockerfile
```diff
@@ -18,8 +18,9 @@ COPY /privbayes/ /SDGym/privbayes
 WORKDIR /SDGym
 
 # Install project
-RUN make install-all compile
-RUN pip install -U numpy==1.20
+RUN pip install ydata-synthetic==0.6.1
+RUN pip install .[gretel] --no-binary pomegranate
+RUN make compile
```
katxiao, line 23: Had a question - was there an issue with running `make install-all compile`?

katxiao (approved):
lgtm, just had a few questions

Dockerfile
csala: The reason is just the combination of dependencies. `ydata-synthetic` depends on `numpy~=1.19`, which is a `numpy` version we support only for `python < 3.8`. Because of this, if we use 3.8 we end up installing a numpy version which ydata would not be officially compatible with. Since the conflicts do not show up on 3.7 and we also support it, I decided to switch to it.

Dockerfile
csala: Yes. The problem is that when running through `make` I did not have the `--no-binary pomegranate` flag, without which the installation failed when building the docker image.
On the other hand, outside of the docker it is not needed, so adding it to the `Makefile` target seemed unnecessary.

Makefile
```diff
@@ -104,9 +104,14 @@ install-gretel-develop: clean-build clean-compile clean-pyc compile ## install t
 
 .PHONY: install-all
 install-all: clean-build clean-compile clean-pyc compile ## install the package with gretel and ydata
-	pip install 'ydata-synthetic>=0.3.0,<0.4'
+	pip install 'ydata-synthetic==0.6.1'
```
kveerama, line 107: We updated ``ydata`` since the older version is no longer working

setup.py
```diff
@@ -43,12 +43,11 @@
 
 ydata_requires = [
     # preferably install using make install-ydata
-    'ydata-synthetic>=0.3.0,<0.4',
+    'ydata-synthetic==0.6.1',
```
kveerama, line 46: We are now pinning them, and updating would be done upon our review of their update.

kveerama: Let's add in documentation, if one wants to update to the updated version of ``ydata`` we would welcome such contribution, and we can review.

katxiao (approved):
LGTM

Issue #134, opened 2022-01-26
Update summary metrics


katxiao:
### Problem Description

Current, the summary metrics from the `sdgym summary` command are `% coverage, avg time, avg score`. We can update them to be more informative about the synthesizers' comparative performance and accuracy.

### Expected behavior

Performance (time) metrics: One idea is to use a leaderboard approach. We can display how many times a given synthesizer had the best relative modeling time.

Quality metrics: TBD

PR #135, opened 2022-01-26
Update the time metrics in summary

katxiao:
Address the time metric updates in #134 

Also remove `Identity` synthesizer from the summary metrics.

### DIFF
--- a/sdgym/summary.py
+++ b/sdgym/summary.py
@@ -72,11 +72,18 @@ def _beat_baseline(data, baseline_scores):
         _synthesizer_beat_baseline, baseline_scores=baseline_scores)
 
 
+def _best_time(data, rank):
+    ranks = data.groupby('dataset')['model_time'].rank('dense')
+    ranks = pd.DataFrame({'rank': ranks, 'synthesizer': data['synthesizer']})
+    ranks = ranks.groupby('synthesizer').apply(lambda x: x[x['rank'] == rank].count())
+    return ranks['rank']
+
+
 def summarize(data, baselines=(), datasets=None):
     """Obtain an overview of the performance of each synthesizer.
 
     Optionally compare the synthesizers with the indicated baselines or analyze
-    only some o the datasets.
+    only some of the datasets.
 
     Args:
         data (pandas.DataFrame):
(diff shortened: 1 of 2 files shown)


sdgym/summary.py
```diff
@@ -72,11 +72,18 @@ def _beat_baseline(data, baseline_scores):
         _synthesizer_beat_baseline, baseline_scores=baseline_scores)
 
 
+def _best_time(data, rank):
```
csala, line 75: I think this could be merged with the `_best` function above by accepting the field name as input, which would make reading easier because developers only need to learn 1 approach.

The function becomes something like this:

```python
def _best(data, rank, field):
    ranks = data.groupby('dataset').rank(method='dense', ascending=False)[field] == rank
    return ranks.groupby(data.synthesizer).sum()
```

In it can be used as:

```python3
        'best': _best(no_identity, 1, 'normalized_score'),
        'best_time': _best_time(no_identity, 1, 'model_time'),
        'second_best_time': _best_time(no_identity, 2, 'model_time'),
        'third_best_time': _best_time(no_identity, 3, 'model_time'),
```

sdgym/summary.py
```diff
@@ -190,12 +201,41 @@ def add_sheet(dfs, name, writer, cell_fmt, index_fmt, header_fmt):
...
+
+
+def _add_summary_categories(summary_data):
+    categories = {
+        'SDV': ['ctgan', 'copulagan', 'gaussiancopula', 'tvae', 'hma1', 'par'],
+        'Gretel': ['gretel'],
+        'YData': ['dragan', 'vanilllagan', 'wgan'],
+    }
+    summary_data['category'] = summary_data.index.map(
+        lambda x: _find_category(x, categories))
```
csala, line 220: The `lambda` is not needed here. If we still need to pass down categories anyway (see coment above), we can use `apply` instead of `map` and pass the `categories` kwarg down directly:

```python3
summary_data['category'] = summary_data.index.apply(_find_category, categories=categories)
```

sdgym/summary.py
```diff
@@ -190,12 +201,41 @@ def add_sheet(dfs, name, writer, cell_fmt, index_fmt, header_fmt):
...
+    for category, category_synthesizers in categories.items():
+        for category_synthesizer in category_synthesizers:
+            if category_synthesizer in synthesizer.lower():
+                return category
+
+    return None
+
+
+def _add_summary_categories(summary_data):
+    categories = {
```
csala, line 214: I think this should be made a module level constant, and then we do not need to pass it as an argument.

csala (changes requested):

csala (approved):
It looks good now!

PR #137, opened 2022-03-29
Import TVAE model

tejuafonja:
TVAE model is not imported alongside the remaining synthesizers. This change fixes that. It imports TVAE model from synthesizers.sdv and also include it to __all__.

PR #140, opened 2022-04-27
Correctly sort imported models

tejuafonja:
We are missing `TVAE`, `HMA1`, and `PAR` models in `sdgym/synthesizers/__init__.py`. #137 fixes this, however, the models were incorrectly sorted, I have fixed that with this change.


katxiao (approved):
Thanks for making the change @tejuafonja! LGTM :shipit:

tejuafonja:
Thanks, @katxiao😎

PR #153, opened 2022-11-01
Update synthesizers to new API

### LINKED ISSUE #154: Updating existing synthesizers to new API
### Problem Description

The existing synthesizers should be updated to use the new synthesizer API, which splits `fit_sample` into two methods - `create_trained_synthesizer` and `sample_synthesizer`.

Also update the synthesizer naming and remove any unused synthesizers.

katxiao:
Resolves #154

- Update synthesizer names
- Remove unneeded synthesizers
- Update all existing synthesizers to the new API


sdgym/synthesizers/base.py
```diff
@@ -59,36 +62,39 @@ class SingleTableBaseline(Baseline, abc.ABC):
...
-            return {
-                table_name: _fit_sample(table, metadata.get_table_meta(table_name))
-                for table_name, table in real_data.items()
-            }
+    def get_trained_synthesizer(self, data, metadata):
+        return self._get_transformed_trained_synthesizer(data, metadata) if (
+            self.CONVERT_TO_NUMERIC) else self._get_trained_synthesizer(data, metadata)
 
-        return _fit_sample(real_data, metadata)
+    def sample_synthesizer(self, synthesizer, n_samples):
```
amontanez24, line 92: can we add docstrings for these methods? Also, I know it's in the issue, but to me `sample_synthesizer` sounds like you are going to sample a new synthesizer as opposed to sampling from it

sdgym/synthesizers/independent.py
```diff
@@ -2,32 +2,34 @@
...
+class IndependentSynthesizer(MultiSingleTableBaselineSynthesizer):
     """Synthesizer that learns each column independently.
 
     Categorical columns are sampled using empirical frequencies.
     Continuous columns are learned and sampled using a GMM.
     """
 
-    @staticmethod
-    def _fit_sample(real_data, metadata):
+    def _get_trained_synthesizer(self, real_data, metadata):
```
amontanez24, line 15: why is this called `get_trained_synthesizer` if it returns metadata and transformed data?

amontanez24 (commented):
This seems to address the issue, but I think that some of the naming in here is confusing

sdgym/synthesizers/independent.py
katxiao: This is kind of a special case synthesizer, because we created one to sample columns independently, and there is no actual model that it's referencing. Instead of making a model class that stores the metadata, etc, I just returned it as a tuple.

amontanez24 (approved):
Thanks for addressing! :shipit:

PR #164, opened 2022-12-15
Create single table benchmarking method

### LINKED ISSUE #151: Create `benchmark_single_table` method
### Problem Description

Create the `benchmark_single_table` method which will benchmark a single table model on the SDV demo datasets. This is very similar to our existing `benchmark` method, but with some API changes.

New functionality:
Instead of only running metrics, we should be able to run reports and include the overall report score in the benchmarking results. Right now, we will add an option to include the Quality Report's overall score.

### Expected behavior
`sdgym.benchmark_single_table`
- `synthesizers`: A list of synthesizer classes to use.
    - (default) `[GaussianCopulaSynthesizer, FASTMLPreset, CTGANSynthesizer]`
    - Options: any of the SDV synthesizers, as available through SDGym
- `sdv_datasets`: A list of strings corresponding to the SDV demo datasets
    - (default) `[adult, alarm, census, child, expedia_hotel_logs, insurance, intrusion, news, covtype]`
    - Options: Various strings, corresponding to single-table datasets
    - Use None to disable using any sdv datasets
- `additional_datasets_folder`: A string containing the path to a folder (local or an S3 bucket). Datasets found in this folder are run in addition to the SDV datasets.
    - (default) `None` – do not run any datasets
- `limit_dataset_size`: Use this flag to limit the size of the datasets for faster evaluation
    - (default) False: Do not limit the size. Use the full dataset
    - `True`: Limit the size of every table to 100 rows (randomly sampled) and the first 10 columns.
- `evaluate_quality`: A boolean representing whether or not to evaluate an overall quality score
    - (default) `True`: Compute an overall data quality score
    - `False`: Do not evaluate data quality. If you set this, only performance and error-related information will be available.
- `sdmetrics`: A list of strings corresponding to the different SDMetrics to use. If you'd like to input specific parameters into the metric, provide a tuple with the metric name followed by a dictionary of the parameters.
    - (default) `[('NewRowSynthesis', {'synthetic_sample_size': 1_000 }]`: Do not run any additional metrics. Only show the basic scores
    - Options: Various (see SDMetrics guide for details)
- `timeout`: The maximum number of seconds to wait for synthetic data creation
    - (default) None: Run for as long as it takes to finish the synthetic data creation
- `output_filepath`: A file path for where to write the output (as a csv file)
    - (default) `None`: Do not write the output anywhere
    - <string>: Write the output to this file name as a CSV file
- `detailed_results_folder`: The folder for where to store the intermediary results
    - (default) `None`: Do not store the intermediary results anywhere
    - <string>: Store the intermediary results in this folder. If the program crashes or takes a long time to run, you can view the results in the folder at any time.
- `show_progress`: Same as before

Returns a pandas.DataFrame with the results from running the benchmark

katxiao:
Resolves #151


sdgym/benchmark.py
```diff
@@ -253,157 +267,139 @@ def _run_on_dask(jobs, verbose):
...
+            A list of the different SDMetrics to use. If you'd like to input specific parameters
+            into the metric, provide a tuple with the metric name followed by a dictionary of
+            the parameters.
+        timeout (bool or ``None``):
+            The maximum number of seconds to wait for synthetic data creation. If ``None``, no
+            timeout is enforced.
+        output_filepath (str or ``None``):
+            A file path for where to write the output as a csv file. If ``None``, no output
+            is written.
+        detailed_results_folder (str or ``None``):
```
amontanez24, line 315: I think this should be called `intermediary_results_folder`. The issue also seems to suggest that

sdgym/benchmark.py
```diff
@@ -253,157 +267,139 @@ def _run_on_dask(jobs, verbose):
...
+    if detailed_results_folder and not is_s3_path(detailed_results_folder):
+        detailed_results_folder = Path(detailed_results_folder)
+        os.makedirs(detailed_results_folder, exist_ok=True)
 
-    run_id = os.getenv('RUN_ID') or str(uuid.uuid4())[:10]
+    max_rows = None
+    max_columns = None
+    if limit_dataset_size:
+        max_rows = 1000
+        max_columns = 10
```
amontanez24, line 339: minor but I think this could be one line
```python
max_rows, max_columns = 1000, 10 if limit_dataset_size else None, None
```

sdgym/benchmark.py
```diff
@@ -253,157 +267,139 @@ def _run_on_dask(jobs, verbose):
...
+                - ``CopulaGANSynthesizer``
+                - ``TVAESynthesizer``
+                - ``FASTMLPreset``
+                - any custom created synthesizer or variant
+
+        sdv_datasets (list[str] or ``None``):
+            Names of the SDV demo datasets to use for the benchmark. Defaults to
+            ``[adult, alarm, census, child, expedia_hotel_logs, insurance, intrusion, news,
+            covtype]``. Use ``None`` to disable using any sdv datasets.
+        additional_datasets_folder (str or ``None``):
```
amontanez24, line 296: Is the functionality for this not being added yet?

sdgym/benchmark.py
```diff
@@ -253,157 +267,139 @@ def _run_on_dask(jobs, verbose):
...
-        for synthesizer, dataset, iteration in jobs:
-            job_tuples.append((
-                get_synthesizers([synthesizer])[0],
-                get_dataset_paths([dataset], datasets_path, bucket, aws_key, aws_secret)[0],
-                iteration
-            ))
+    job_tuples = list()
+    for dataset in datasets:
+        for synthesizer in synthesizers:
+            job_tuples.append((synthesizer, dataset, 1))
```
amontanez24, line 349: any reason not to filter for modalities here? Then you wouldn't need an if statement on line 356

sdgym/benchmark.py
katxiao: Ah the issue is out of date. I'll correct it there.

sdgym/benchmark.py
katxiao: Not yet. I'll add it in a following PR

sdgym/benchmark.py
katxiao: One reason is so that we won't need to call `load_dataset` in two places.

sdgym/benchmark.py
amontanez24: Is there a reason we changed it? Intermediary seems to make more sense to me

amontanez24 (approved):
Just one question about a variable name but otherwise looks good!

sdgym/benchmark.py
katxiao: I think `intermediate` seemed like it would require the user to understand more about the implementation (these results are collected throughout the run), and `detailed_results_folder` is just a description of what the folder contains.

PR #165, opened 2022-12-16
Maintenance updates

### LINKED ISSUE #163: SDGym Package Maintenance Updates
_This is a placeholder issue for overall package maintenance updates for the next minor SDGym release: `0.6.0`_

katxiao:
Resolves #163

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,108 @@
-MIT License
+Business Source License 1.1
 
-Copyright (c) 2019, MIT Data To AI Lab
+Parameters
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
-of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
-to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
-copies of the Software, and to permit persons to whom the Software is
-furnished to do so, subject to the following conditions:
+Licensor:             DataCebo, Inc.
 
-The above copyright notice and this permission notice shall be included in all
-copies or substantial portions of the Software.
+Licensed Work:        SDGym
+                      The Licensed Work is (c) DataCebo, Inc.
 
-THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
-IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
-FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
-AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
-LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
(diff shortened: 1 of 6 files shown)


LICENSE
```diff
@@ -1,21 +1,108 @@
...
+
+Licensed Work:        SDGym
+                      The Licensed Work is (c) DataCebo, Inc.
+
+Additional Use Grant: You may make use of the Licensed Work, and derivatives of the Licensed
+                      Work, provided that you do not use the Licensed Work, or derivatives of
+                      the Licensed Work, for a Synthetic Data Benchmarking Service.
+
+                      A "Synthetic Data Benchmarking Service" is a commercial offering
+                     	that allows third parties (other than your employees and
```
npatki, line 15: Formatting seems off. There should be the same # of spaces as line 14

LICENSE
```diff
@@ -1,21 +1,108 @@
...
+                     	contractors) to access the functionality of the Licensed
```
npatki, line 16: Formatting seems off. There should be the same # of spaces as line 14

LICENSE
```diff
@@ -1,21 +1,108 @@
...
+                      Work so that such third parties directly benefit from the
+                      data specification, data transformation, machine learning, synthetic
+                      data creation, synthetic data evaluation or performance evaluation
+                      features of the Licensed Work.
+
+
+Change Date:          Change date is four years from release date.
+Please see https://github.com/sdv-dev/SDGym/releases
```
npatki, line 24: This should spaced so it starts with the "Change date is four years from release date." sentence above

README.md
```diff
@@ -22,163 +23,70 @@
 
 # Overview
```
npatki, line 24: As soon as this is merged into `master`, the README will update. For this PR I would suggest:

1. Keeping the same README as before and
2. Just removing the row in the table that says `LICENSE`, so that it's not conflicting with the changes

image

npatki (changes requested):

sdgym/__init__.py
```diff
@@ -4,10 +4,10 @@
...
 """
 
-__author__ = 'MIT Data To AI Lab'
-__copyright__ = 'Copyright (c) 2018, MIT Data To AI Lab'
-__email__ = 'dailabmit@gmail.com'
-__license__ = 'MIT'
+__author__ = 'DataCebo, Inc.'
+__copyright__ = 'Copyright (c) 2022 DataCebo, Inc.'
+__email__ = 'info@sdv.dev'
+__license__ = 'BSL-1.1'
```
pvk-developer, line 10: I think this is not added on the other libraries, should we add it? `license` and `copyright` @npatki

pvk-developer (approved):

sdgym/__init__.py
npatki: I'm curious .. where does this info ultimately show up? 

We probably should add it for all libraries, but I'd generally consider it non-blocking (pending the Q above).

LICENSE
npatki: I'm still seeing an issue with this @katxiao -- am I looking at the wrong version? The intended formatting is to add spaces before "Please see" to keep it inline. See image below.

image

npatki (changes requested):
One small formatting change in the LICENSE file

npatki (approved):
LGTM!

PR #167, opened 2022-12-22
Allow additional datasets folder

### LINKED ISSUE #166: Allow custom datasets
### Problem Description

Allow SDGym to run on custom datasets by letting the user to specify a local folder or s3 bucket to read datasets from. The folder should have the expected structure
- SINGLE_TABLE/
- MULTI_TABLE/
- TIMESERIES/

### Expected behavior

All datasets of the requested modality are run in addition to the specified SDV datasets.

katxiao:
Resolves #166

### DIFF
--- a/sdgym/datasets.py
+++ b/sdgym/datasets.py
@@ -46,6 +46,10 @@ def _get_dataset_path(modality, dataset, datasets_path, bucket=None, aws_key=Non
     if dataset_path.exists():
         return dataset_path
 
+    local_path = Path(bucket) / dataset
+    if local_path.exists():
+        return local_path
+
     download_dataset(
         modality, dataset, dataset_path, bucket=bucket, aws_key=aws_key, aws_secret=aws_secret)
     return dataset_path
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -342,6 +342,9 @@ def benchmark_single_table(synthesizers=DEFAULT_SYNTHESIZERS, sdv_datasets=DEFAU
 
     synthesizers = get_synthesizers(synthesizers)
     datasets = get_dataset_paths(sdv_datasets, None, None, None, None)
+    if additional_datasets_folder:
+        additional_datasets = get_dataset_paths(None, None, additional_datasets_folder, None, None)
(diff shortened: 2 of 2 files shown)


sdgym/datasets.py
```diff
@@ -46,7 +46,7 @@ def _get_dataset_path(modality, dataset, datasets_path, bucket=None, aws_key=Non
     if dataset_path.exists():
         return dataset_path
 
-    local_path = Path(bucket) / dataset
+    local_path = Path(bucket) / dataset if bucket else Path(dataset)
```
amontanez24, line 49: does this work on all operating systems or should we us `os.join`?

sdgym/synthesizers/generate.py
```diff
@@ -147,7 +148,7 @@ def create_multi_table_synthesizer(get_trained_synthesizer_fn, sample_from_synth
         class:
             The synthesizer class.
     """
-    class NewSynthesizer(BaselineSynthesizer):
+    class NewSynthesizer(MultiSingleTableBaselineSynthesizer):
```
amontanez24, line 151: just a question but why is it called `MultiSingleTable` instead of just `MultiTable`?

amontanez24 (commented):
Looks good! Just a couple of questions

sdgym/datasets.py
katxiao: I'm not sure, since I haven't tested it out for Windows, but I see your concern. I'll just switch it to using `os.path.join`.

sdgym/synthesizers/generate.py
katxiao: We use `MultiSingleTable` in SDGym and SDMetrics. It already existed in the repos so I was just maintaining consistency, but we can discuss the naming further offline.

sdgym/datasets.py
katxiao: Actually it seems like using path operators should be compatible with all platforms since it creates a new `Path` object (not just a string) -https://docs.python.org/3/library/pathlib.html#operators

amontanez24 (approved):

PR #178, opened 2023-01-18
Catch errors in benchmarking

### LINKED ISSUE #177: Benchmarking error handling
### Problem Description

If a benchmarking run fails for a (synthesizer, dataset) combination, we should catch the error and proceed to try benchmarking all (synthesizer, dataset) combinations. For the failed runs, we can display `NaN` in the results table for those rows. If a `detailed_results_folder` is provided, we can write the error to the `detailed_results_folder`.

### Expected behavior

If there are errors, we should display `NaN` in the results table for the entries that were unable to be calculated.

image

katxiao:
Resolves #177


fealho (approved):
Pretty clean 👍

PR #185, opened 2023-01-19
Add s3 prefix and fix get size

### LINKED ISSUE #182: Add option to disable sdmetrics in benchmarking
### Problem Description

The benchmarking method should be able to take in an empty list of empty metrics. Right now `sdmetrics=None` is how you disable metrics from running. Passing in an empty list should have the same effect.

### Expected behavior
The following call should not run any metrics:
```
benchmark_single_table(
    ...
    sdmetrics=[],
)
```

### LINKED ISSUE #183: Prefix remote bucket with 's3'
### Problem Description

If the user supplies an S3 bucket, we should have them prefix the URL with "s3://"

Eg. "s3://sdv-demo-datasets"

### Expected behavior

The following command should run the datasets in the `sdv-demo-datasets` s3 bucket.

```
benchmark_single_table(
    ...
    additional_datasets_folder='s3://sdv-demo-datasets',
)
```

### LINKED ISSUE #184: Size values in benchmarking results seems inaccurate
### Environment Details

The sizes for dataset size and synthesizer size (in MB) seems to be very low. 

For small datasets, it's reporting model size to be 0.046KB. It seems a couple orders of magnitude off. We should double check that we are computing the size of these objects correctly.

katxiao:
Resolves #184 
Resolves #183 
Resolves #182

### DIFF
--- a/sdgym/datasets.py
+++ b/sdgym/datasets.py
@@ -133,7 +133,9 @@ def get_available_datasets(modality, bucket=None, aws_key=None, aws_secret=None)
 
     s3 = get_s3_client(aws_key, aws_secret)
     bucket = bucket or BUCKET
-    bucket_name = bucket[len(S3_PREFIX):]
+    if bucket.startswith(S3_PREFIX):
+        bucket_name = bucket[len(S3_PREFIX):]
+
     response = s3.list_objects(Bucket=bucket_name, Prefix=modality.upper())
     datasets = []
     for content in response['Contents']:
--- a/sdgym/datasets.py
+++ b/sdgym/datasets.py
@@ -21,11 +21,15 @@ MODALITIES = ['single_table', 'multi_table', 'sequential']
 S3_PREFIX = 's3://'
 
 
+def _get_bucket_name(bucket):
+    return bucket[len(S3_PREFIX):] if bucket.startswith(S3_PREFIX) else bucket
+
+
 def download_dataset(modality, dataset_name, datasets_path=None, bucket=None, aws_key=None,
                      aws_secret=None):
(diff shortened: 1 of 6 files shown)


pvk-developer (approved):

sdgym/utils.py
```diff
@@ -230,3 +231,38 @@ def _synthesizer_sample_function(instance, n_samples=None):
...
+        seen (set):
+            The ids of the objects that have already been evaluated.
+
+    Returns:
+        int:
+            The size in bytes.
+    """
+    size = sys.getsizeof(obj)
+    if seen is None:
+        seen = set()
```
pvk-developer, line 251: Just a nit, could we add a break line ?

sdgym/datasets.py
```diff
@@ -14,20 +14,22 @@
...
 TIMESERIES_FIELDS = ['sequence_index', 'entity_columns', 'context_columns', 'deepecho_version']
 MODALITIES = ['single_table', 'multi_table', 'sequential']
+S3_PREFIX = 's3://'
 
 
 def download_dataset(modality, dataset_name, datasets_path=None, bucket=None, aws_key=None,
                      aws_secret=None):
     datasets_path = datasets_path or DATASETS_PATH / dataset_name
     bucket = bucket or BUCKET
+    bucket_name = bucket[len(S3_PREFIX):] if bucket.startswith(S3_PREFIX) else bucket
```
amontanez24, line 28: minor but we can make this a function since you use it later too. Like a `get_bucket_name_from_path` of something

amontanez24 (changes requested):

amontanez24 (approved):
Thanks for addressing! :shipit:

PR #198, opened 2023-01-24
Update benchmarking and datasets API

### LINKED ISSUE #194: Accept `metadata.json` as valid metadata file.
### Problem Description

Currently, we expect the metadata file in datasets to be named `metadata_v0.json`. This can be confusing since it implies that metadata is versioned. We should accept `metadata.json` as a valid metadata file.

### LINKED ISSUE #195: Rename benchmarking argument "evaluate_quality" to "compute_quality_score"
### Problem Description

Rename argument "evaluate_quality" to "compute_quality_score" in `benchmark_single_table`. The original argument is confusing, since we still evaluate quality metrics when `evaluate_quality` is `False`.

### LINKED ISSUE #196: Check if file or folder exists before writing benchmarking results
### Problem Description

Currently, in the benchmarking method, we can write results to an output file and a detailed results folder if the user specifies these arguments. If the folder or file already exists, it will be silently overwritten by the new benchmarking run.

We should throw an Error when the file or folder already exists, so that the data there isn't unintentionally overwritten. We expect that the file or folder in the arguments to not previously exist.

### LINKED ISSUE #197: Remove arguments in `get_available_datasets`
### Problem Description

For the `get_available_datasets` function,

1. Hardcode the bucket name and remove it as a parameter
2. Hardcode modality='single_table' and also remove it as a parameter, since we only support single table benchmarking right now.

katxiao:
Resolves #195 
Resolves #196 
Resolves #197 
Resolves #194

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -112,7 +112,7 @@ def _prepare_metric_args(real_data, synthetic_data, metadata):
     return real_data, synthetic_data, metadata
 
 
-def _compute_scores(metrics, real_data, synthetic_data, metadata, output, evaluate_quality):
+def _compute_scores(metrics, real_data, synthetic_data, metadata, output, compute_quality_score):
     metrics = metrics or []
     if len(metrics) > 0:
         metrics, metric_kwargs = get_metrics(metrics, metadata)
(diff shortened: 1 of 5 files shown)


fealho (approved):
👍

PR #199, opened 2023-01-25
Use silent report generation in benchmarking

### LINKED ISSUE #179: Use SDMetrics silent report generation
### Problem Description

Use the new silent report generation available in SDMetrics 0.9.0.

### Expected behavior

When we are benchmarking, we should no longer see the summary and progress bar of SDMetrics report generation.

### LINKED ISSUE #192: `IndependentSynthesizer` is printing out `ConvergeWarning` too many times
### Problem Description
When applying the `IndependentSynthesizer` to demo data (eg. `student_placements_pii`), I see that it is printing out `ConvergenceWarning` many times.

```
ConvergenceWarning: Number of distinct clusters (2) found smaller than n_clusters (5). Possibly due to duplicate points in X.
ConvergenceWarning: Number of distinct clusters (3) found smaller than n_clusters (5). Possibly due to duplicate points in X.
ConvergenceWarning: Number of distinct clusters (1) found smaller than n_clusters (5). Possibly due to duplicate points in X.
etc.
```

### Expected behavior
The root cause is that we attempt to create a `GaussianMixture` with 5 clusters by default -- see [line](https://github.com/sdv-dev/SDGym/blob/master/sdgym/synthesizers/independent.py#L30).

If there are fewer than 5 unique data points, then we should lower the number of clusters.

```python
num_components = min(num_unique_values, 5)
model = GaussianMixture(num_components)
```

### Additional context

We may also need to investigate other other edge cases that cause these warnings.

katxiao:
Resolves #179 
Resolves #192

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -158,10 +158,11 @@ def _compute_scores(metrics, real_data, synthetic_data, metadata, output, evalua
             table_synthetic_data = list(synthetic_data.values())[0]
 
             quality_report = SingleTableQualityReport()
-            quality_report.generate(table_real_data, table_synthetic_data, table_metadata)
+            quality_report.generate(
+                table_real_data, table_synthetic_data, table_metadata, verbose=False)
         else:
             quality_report = MultiTableQualityReport()
-            quality_report.generate(real_data, synthetic_data, metadata)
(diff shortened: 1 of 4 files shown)


pvk-developer (approved):
LGTM!

PR #202, opened 2023-01-26
Move fitting logic to training method

### LINKED ISSUE #193: Training vs. sampling logic for `IndependentSynthesizer` is not fully abstracted
**Version**: 0.8.0 (under development)

### Current
We expect that all synthesizers should cleanly define:

1. A training function that produces a fully trained synthesizer
2. A sampling functions that uses the synthesizer to sample

These abstractions are not being followed in the [IndependentSynthesizer](https://github.com/sdv-dev/SDGym/blob/master/sdgym/synthesizers/independent.py#L8).

- **Training**: The training function is only returning the metadata and the transformed data (preprocessed with RDT)
- **Sampling**: The sampling function is fitting an `sklearn.mixture.GaussianMixture` model to each continuous column before sampling

### Expected

The training phase should include all the necessary information needed to sample. There should be no fitting happening in the sampling call.

- **Training**: We should be fitting & storing a  `GaussianMixture` on each continuous column _and_ storing the unique values of each categorical column during this phase.
- **Sampling**: At this point, we only need to sample from the `GaussianMixture` and unique values to sample.

katxiao:
Resolves #193

### DIFF
--- a/sdgym/synthesizers/independent.py
+++ b/sdgym/synthesizers/independent.py
@@ -38,7 +38,6 @@ class IndependentSynthesizer(MultiSingleTableBaselineSynthesizer):
                 values = column.sample(self.length, replace=True).values
             else:
                 model = gm_models.get(name)
-                model.fit(column.values.reshape(-1, 1))
                 values = model.sample(self.length)[0].ravel().clip(column.min(), column.max())
 
             sampled[name] = values
--- a/sdgym/synthesizers/independent.py
+++ b/sdgym/synthesizers/independent.py
@@ -17,18 +17,27 @@ class IndependentSynthesizer(MultiSingleTableBaselineSynthesizer):
         metadata.fit(real_data)
         transformed = metadata.transform(real_data)
         self.length = len(real_data)
-        return (metadata, transformed)
+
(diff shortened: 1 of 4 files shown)


sdgym/synthesizers/independent.py
```diff
@@ -17,18 +17,27 @@ def _get_trained_synthesizer(self, real_data, metadata):
...
         sampled = pd.DataFrame()
         for name, column in transformed.items():
             kind = column.dtype.kind
             if kind == 'O':
                 values = column.sample(self.length, replace=True).values
             else:
-                num_components = min(column.nunique(), 5)
-                model = GaussianMixture(num_components)
+                model = gm_models.get(name)
                 model.fit(column.values.reshape(-1, 1))
```
amontanez24, line 41: Does it still need to fit here?

amontanez24 (commented):
Just one question, otherwise looks good!

sdgym/synthesizers/independent.py
katxiao: Good call - nope I removed it.

amontanez24 (approved):

PR #208, opened 2023-02-21
Sdgym script

fealho:


sdgym/summary.py
```diff
@@ -268,8 +270,13 @@ def _add_summary(data, modality, baselines, writer):
     add_sheet(performance, f'Performance ({modality})', writer, cell_fmt, index_fmt, header_fmt)
     add_sheet(error_summary, f'Errors Summary ({modality})', writer, cell_fmt, index_fmt,
               header_fmt)
```
frances-h, line 272: Small nitpick, but I think if the call below is being split out into multiple lines then this call should be as well.

sdgym/summary.py
```diff
@@ -95,46 +91,49 @@ def summarize(data, baselines=(), datasets=None):
...
         'time': _seconds(data),
-        'best': _best(no_identity, 1, 'normalized_score', False),
+        'best': _best(no_identity, 1, 'Quality_Score', False),
         'avg score': _mean_score(data),
-        'best_time': _best(no_identity, 1, 'model_time', True),
-        'second_best_time': _best(no_identity, 2, 'model_time', True),
-        'third_best_time': _best(no_identity, 3, 'model_time', True),
+        'best_time': _best(no_identity, 1, 'Train_Time', True),
+        'second_best_time': _best(no_identity, 2, 'Train_Time', True),
+        'third_best_time': _best(no_identity, 3, 'Train_Time', True),
```
amontanez24, line 113: @npatki Are these still the desired columns to get in the output or are some of these unnecessary now?

npatki: @amontanez24 we haven't really discussed what the final output summary should be. In any case, it would change because the underlying benchmarking method has also changed. So it seems fine to get rid of.

pvk-developer (approved):
I think is good 👍🏻

amontanez24 (approved):

PR #213, opened 2023-03-31
Update SDGym to work with SDV 1.0.0

### LINKED ISSUE #212: Cleanup SDGym to the new SDV 1.0 metadata and synthesizers
### Problem Description
Since we've upgraded to [SDV 1.0](https://docs.sdv.dev/sdv), the synthesizer class names and metadata formats have changed. The SDGym is currently written assuming that SDV < 1.0 (even though the official requirements are anything `>=0.18.0`). 

We should fix this asap.

### Expected behavior
Use the synthesizers, datasets and metadata as provided by the SDV 1.0 library. 

We should also update the minimum and maximum to: `sdv >1.0` and `sdv < 2.0`

fealho:
Resolve #212.

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -95,7 +95,7 @@ def _synthesize(synthesizer_dict, real_data, metadata):
 
 
 def _compute_scores(metrics, real_data, synthetic_data, metadata,
-                    output, compute_quality_score, modality):
+                    output, compute_quality_score, modality, dataset_name):
     metrics = metrics or []
     if len(metrics) > 0:
         metrics, metric_kwargs = get_metrics(metrics, metadata, modality='single-table')
@@ -113,10 +113,13 @@ def _compute_scores(metrics, real_data, synthetic_data, metadata,
             normalized_score = None
             start = datetime.utcnow()
             try:
+                LOGGER.info('Computing %s on dataset %s', metric_name, dataset_name)
                 metric_args = (real_data, synthetic_data, metadata)
                 score = metric.compute(*metric_args, **metric_kwargs.get(metric_name, {}))
                 normalized_score = metric.normalize(score)
             except Exception:
+                LOGGER.exception(
+                    'Metric %s failed on dataset %s. Skipping.', metric_name, dataset_name)
                 _, error = format_exception()
 
             scores[-1].update({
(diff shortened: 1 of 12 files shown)


sdgym/benchmark.py
```diff
@@ -133,12 +113,10 @@ def _compute_scores(metrics, real_data, synthetic_data, metadata, output, comput
             normalized_score = None
             start = datetime.utcnow()
             try:
-                LOGGER.info('Computing %s on dataset %s', metric_name, metadata._metadata['name'])
```
amontanez24, line 136: I think we can keep the logs

sdgym/benchmark.py
```diff
@@ -151,75 +129,52 @@ def _compute_scores(metrics, real_data, synthetic_data, metadata, output, comput
 
     if compute_quality_score:
         start = datetime.utcnow()
-        if metadata.modality == 'single-table':
```
amontanez24, line 154: are we getting rid of the modality checks because this PR will drop support for other modalities?

fealho: In this PR I tried to make the minimal changes to make the code work. I didn't want to completely delete the support for other modalities here. I think if we want to do that it should be done in a different PR.

Another option is to not merge this PR to master, and instead to some other branch `ready-for-release`, which in turn will be merged to master. We can then make all the other incremental changes there, if you think merging code with pieces that don't work is a problem.

sdgym/benchmark.py
fealho: I deleted them because we don't have the parameters anymore in most of them. I can try to fix that though.

amontanez24: yeah let's try to get them or something close. I think we could use more logging in our libraries lol

sdgym/synthesizers/base.py
```diff
@@ -146,6 +146,8 @@ class MultiSingleTableBaselineSynthesizer(BaselineSynthesizer, abc.ABC):
 
     These classes model and sample each table independently and then just
     randomly choose ids from the parent tables to form the relationships.
+
+    NOTE: doesn't currently work.
```
pvk-developer, line 150: Out of curiosity, did it work before ? If so do we know why is not currently working ?

pvk-developer (approved):
LGTM! Solid work 🪨

amontanez24 (approved):
LGTM! Thanks for the changes :shipit:

Issue #215, opened 2023-04-07
Code Refactoring


fealho:
### Description

SDGym requires some refactoring. In part due to issue #212, parts of the code are currently misplaced, unused or outdated.

That should be fixed in the following parts:

- Create a `_benchmark` method which all `benchmark_{modality}` methods will call (add helper methods)
- Delete the `MODALITIES` attribute
- Delete `SingleTableBaselineSynthesizer`
- Fix `MultiSingleTableBaselineSynthesizer`
- Move all outdated code into a Legacy folder
- Move all CLI-only code into a CLI folder
- (Maybe) Convert all metadata objects into `SingleTableMetadata`/`MultiTableMetadata`
- `dataset.py` logic should import from SDV
- Remove `SynthesizerMapping` and rename `SingleTablePreset` to `FastMLPreset`
- Add integration tests for all public methods
- Review and update any incorrect docstrings/READMEs
- Other minor fixes

These should all be merged into a separate branch, which will in turn be merged to master.

PR #220, opened 2023-04-19
Remove `MODALITIES` and `privbn` artifacts

fealho:
As part of Issue #215, this PR deletes:
- The `MODALITIES` attribute. We currently don't use it. If it is required for the `multi_table` and `time-series` cases, it should be added back when those classes are implemented.
- `privbn` artifacts which should've been removed in the PR #218.


sdgym/datasets.py
```diff
@@ -122,11 +121,6 @@ def get_available_datasets():
 
 
 def _get_available_datasets(modality, bucket=None, aws_key=None, aws_secret=None):
-    if modality not in MODALITIES:
```
amontanez24, line 125: Should we still have a check to make sure `modality` isn't some random string?

fealho: Very good catch, I shouldn't have deleted `MODALITIES` in this file, it's unrelated to the one in `synthesizers/base.py`.

amontanez24 (approved):
Thanks for addressing!

pvk-developer (approved):
🚀 LGTM!

PR #221, opened 2023-04-21
Update `SingleTableBaselineSynthesizer`

fealho:
As part of #215, update the `SingleTableBaselineSynthesizer` class. The `_get_transformed_trained_synthesizer` and `_get_reverse_transformed_samples` methods were only used with Gretel, which we removed support for half a year ago.


sdgym/synthesizers/base.py
```diff
@@ -65,74 +64,7 @@ def sample_from_synthesizer(synthesizer, n_samples):
...
-            n_samples (int):
-                The number of samples to create.
-
-        Returns:
-            pandas.DataFrame:
-                The sampled data.
-        """
-        return self._get_reverse_transformed_samples(synthesizer, n_samples) if (
-            self.CONVERT_TO_NUMERIC) else self._sample_from_synthesizer(synthesizer, n_samples)
+        return self._sample_from_synthesizer(synthesizer, n_samples)
```
amontanez24, line 67: I think this was originally just meant to be an abstract method which is why it's empty. We should just raise a not implemented error here. Some classes don't write a `_sample_from_synthesizer` and just override this directly

amontanez24 (changes requested):

sdgym/synthesizers/base.py
fealho: Sure. I implemented this way since that's what we had discussed, but I can make the changes.

amontanez24: sorry, I misread all this. you're right, it's fine to get rid of the `SingleTableBaselineSynthesizer`

amontanez24 (approved):
LGTM!

PR #222, opened 2023-04-21
Add helper methods to `benchmark_single_table` and remove `SYNTHESIZER_MAPPING`

fealho:
As part of #215, this PR:
- Restructures the `benchmark_single_table` function, creating helper methods to make the code easier to understand.
- Removes `SYNTHESIZER_MAPPING`, which used to be useful when the synthesizers in SDV and SDGym had different names. Since they have the same name now, this dictionary is no longer necessary.


pvk-developer (approved):
LGTM!

amontanez24 (approved):
Nice job! :shipit:

PR #224, opened 2023-04-26
Add integration tests

fealho:
As part of #215, add integration tests for the user facing methods. These are: `benchmark_single_table`, `get_available_datasets`, `create_sdv_synthesizer_variant` and `create_single_table_synthesizer`.

### DIFF
--- a/sdgym/synthesizers/generate.py
+++ b/sdgym/synthesizers/generate.py
@@ -5,6 +5,9 @@ from sdv.multi_table.hma import HMASynthesizer
 from sdv.sequential import PARSynthesizer
 from sdv.single_table import (
     CopulaGANSynthesizer, CTGANSynthesizer, GaussianCopulaSynthesizer, TVAESynthesizer)
+from sdv.metadata.single_table import SingleTableMetadata
+from sdv.metadata.multi_table import MultiTableMetadata
+
 
 from sdgym.synthesizers.base import (
     BaselineSynthesizer, MultiSingleTableBaselineSynthesizer, SingleTableBaselineSynthesizer)
@@ -125,6 +128,7 @@ def create_single_table_synthesizer(display_name, get_trained_synthesizer_fn,
                 obj:
                     The trained synthesizer.
             """
+            metadata = SingleTableMetadata().load_from_dict(metadata)
             return get_trained_synthesizer_fn(data, metadata)
 
         def sample_from_synthesizer(self, synthesizer, num_samples):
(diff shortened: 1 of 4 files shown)


sdgym/synthesizers/generate.py
```diff
@@ -125,6 +127,7 @@ def get_trained_synthesizer(self, data, metadata):
                 obj:
                     The trained synthesizer.
             """
+            metadata = SingleTableMetadata().load_from_dict(metadata)
```
fealho, line 130: Bug the new test caught. The metadata is being passed as a dictionary, but  `get_trained_synthesizer_fn` expects the actual `Metadata` object.

I'll change these in another PR, where I'll pass the Metadata directly in the `benchmark_single_table` function, I just didn't want to do it here so it doesn't become too messy.

fealho: It was decided this should be kept as is, since for these methods it is easier for the user to implement the metadata as a dictionary directly.

tests/integration/test_benchmark.py
```diff
@@ -112,3 +115,82 @@ def test_duplicate_synthesizers():
...
+    assert set(results['Dataset_Size_MB']) == {0.026358}
+    assert results['Train_Time'].between(0, 1000).all()
+    assert results['Peak_Memory_MB'].between(0, 100).all()
+    assert results['Synthesizer_Size_MB'].between(0, 100).all()
+    assert results['Sample_Time'].between(0, 100).all()
+    assert results['Evaluate_Time'].between(0, 100).all()
+    assert results['Quality_Score'].between(.6, 1).all()
+    assert results['NewRowSynthesis'][4] == 0
+
+    results['NewRowSynthesis'][4] = 1
```
amontanez24, line 195: Just out of curiosity, why is the 4th value different?

amontanez24 (approved):
One question but otherwise looks good 👍

tests/integration/test_benchmark.py
fealho: Because all values generated by the IdentitySynthesizer match the original data, so there are no new rows.

amontanez24: In that case, minor nitpick but could we make a variable to store `results['NewRowSynthesis'][4]` as like `identity_new_row_synthesis_value` or something so it's clear what's being evaluated here?

pvk-developer (approved):
LGTM!

PR #225, opened 2023-04-27
Fix `UniformSynthesizer` and `IndependentSynthesizer`

fealho:
As part of #215, update `UniformSynthesizer` and  `IndependentSynthesizer` so they work as intended.


pvk-developer (approved):
:shipit:

sdgym/synthesizers/uniform.py
```diff
@@ -24,8 +28,6 @@ def _sample_from_synthesizer(self, synthesizer, n_samples):
             kind = column.dtype.kind
             if kind == 'i':
                 values = np.random.randint(column.min(), column.max() + 1, size=self.length)
-            elif kind == 'O':
-                values = np.random.choice(column.unique(), size=self.length)
```
amontanez24, line 28: why did we get rid of this?

sdgym/synthesizers/independent.py
```diff
@@ -8,38 +8,30 @@
...
 
-        self.length = len(real_data)
         gm_models = {}
         for name, column in transformed.items():
-            kind = column.dtype.kind
-            if kind != 'O':
-                num_components = min(column.nunique(), 5)
-                model = GaussianMixture(num_components)
-                model.fit(column.values.reshape(-1, 1))
-                gm_models[name] = model
```
amontanez24, line 29: what's the motivation to change the logic for sampling categoricals?

sdgym/synthesizers/uniform.py
fealho: The old code allowed us to not pass any transforms to an object - so would be analogous to doing `update_transformers_by_sdtype('categorical', transformer_name=None)` - which we can't do anymore. This distinction is why I had to make some changes to both synthesizers.

In this case, the `fit` always transforms all columns into numerical, so this line never runs, which is why I deleted it. That's the same reason I removed the check in line 37 for the `IndependentSynthesizer`.

PR #235, opened 2023-05-25
Update top level import functions

### LINKED ISSUE #229: Add functions to top level import
### Problem Description

_As.a user, it's easier to import certain functions directly from sdgym._

### Expected behavior

- You should be able to do the following without problem
```python
from sdgym import get_available_datasets
from sdgym import create_sdv_synthesizer_variant
```

### Additional context

- Make sure all other functions that should be importable from sdgym are

fealho:
Resolve #229.


amontanez24 (approved):
:shipit:

PR #236, opened 2023-05-29
Remove code support for synthesizers that are not strings/classes

fealho:
Remove support for synthesizers to be passed as methods/functions/tuples/dictionaries, as well as some minor refactoring of the code.

NOTE: this is an internal change. The refactoring only removes code artifacts, and shouldn't affect the user facing API.

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -80,7 +80,6 @@ def _generate_job_args_list(limit_dataset_size, sdv_datasets, additional_dataset
     run_id = os.getenv('RUN_ID') or str(uuid.uuid4())[:10]
 
     # Get list of synthesizer objects
-    synthesizers = DEFAULT_SYNTHESIZERS if synthesizers is None else synthesizers
     custom_synthesizers = [] if custom_synthesizers is None else custom_synthesizers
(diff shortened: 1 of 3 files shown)


sdgym/benchmark.py
```diff
@@ -80,18 +79,15 @@ def _generate_job_args_list(limit_dataset_size, sdv_datasets, additional_dataset
...
-    synthesizers = get_synthesizers(synthesizers)
-    if custom_synthesizers:
-        custom_synthesizers = get_synthesizers(custom_synthesizers)
-        synthesizers.extend(custom_synthesizers)
-
-    datasets = []
-    if sdv_datasets is not None:
-        datasets = get_dataset_paths(sdv_datasets, None, None, None, None)
+    # Get list of synthesizer objects
+    custom_synthesizers = [] if custom_synthesizers is None else custom_synthesizers
```
pvk-developer, line 83: small nit pick you can leave it as it is if you like it more, but usually:
```python
custom_synthesizers = custom_synthesizers if custom_synthesizers else []
```
In this case its not a huge difference but in the following ones you see what method is going to be called or what they should be if that is true, otherwise a quick glance tells you that this is empty list.

pvk-developer (approved):
LGTM!

PR #239, opened 2023-05-29
Fix timeout argument causing `benchmark_single_table` to crash

### LINKED ISSUE #234: timeout parameter causes sdgym to crash
### Environment Details

* SDGym version: master
* Python version: Any
* Operating System: Any

### Error Description

If the timeout parameter is provided, `benchmark_single_table` crashes with the following error:
```bash
AttributeError: 'dict' object has no attribute 'dropna'
AttributeError: 'list' object has no attribute 'to_dict'
Error running DataIdentity on dataset None;
```
Additionally, the docstring for timeout is incorrect. It says the type is bool when it should be an int.

### Steps to reproduce

``` python
benchmark_single_table(
    synthesizers=['DataIdentity', 'IndependentSynthesizer', 'UniformSynthesizer'],
    sdv_datasets=['student_placements'],
    timeout=10,
)
```

pvk-developer:
Resolves #234 

PS: We should check if `timeout` works for windows.


sdgym/benchmark.py
```diff
@@ -352,6 +355,7 @@ def _run_job(args):
             output = _score_with_timeout(
                 timeout,
                 synthesizer,
+                data,
                 metadata,
                 metrics,
                 max_rows=max_rows,
```
amontanez24, line 361: can we add the kwargs to the top few parameters? ie. 
```python
timeout=timeout,
synthesizer=synthesizer
...
```

tests/unit/test_benchmark.py
```diff
@@ -42,3 +42,35 @@ def test_progress_bar_updates(tqdm_mock):
 
     # Assert
     tqdm_mock.assert_called_once_with(ANY, total=1, position=0, leave=True)
+
+
+@patch('sdgym.benchmark._score_with_timeout')
```
amontanez24, line 47: we don't test `_score_with_timeout` at all. Instead of mocking that, can we just mock the `process` calls and make sure the timeout is passed correctly? You don't have to write tests for `_score_with_timeout`, but just test the functionality here.
https://github.com/sdv-dev/SDGym/blob/bd5693c3d0aabf0e862f6532ff51105d220622fd/sdgym/benchmark.py#L286-L288

tests/integration/test_benchmark.py
```diff
@@ -195,3 +196,32 @@ def sample_from_synthesizer(synthesizer, n_samples):
...
+    # Run
+    scores = sdgym.benchmark_single_table(
+        synthesizers=['CTGANSynthesizer'],
+        sdv_datasets=['student_placements'],
+        timeout=5
+    )
+    total_time = time.time() - start_time
+
+    # Assert
+    assert total_time < 10.0
```
amontanez24, line 214: Is this large enough? No reason to fail this test because it takes 11 seconds

amontanez24 (changes requested):

tests/integration/test_benchmark.py
pvk-developer: This should be good yeah, we want it to timeout as early as possible. Timeout is set to 5 and this is 2x that.

tests/unit/test_benchmark.py
```diff
@@ -47,56 +47,37 @@ def test_progress_bar_updates(tqdm_mock):
 @patch('sdgym.benchmark._score')
 @patch('sdgym.benchmark.multiprocessing')
 def test__score_with_timeout(mock_multiprocessing, mock__score):
```
amontanez24, line 49: minor: rename for the function being called (ie. `benchmark_single_table`)

amontanez24 (approved):
One small comment but after that :shipit:

fealho:
@pvk-developer I removed my approval until the tests are passing.

PR #240, opened 2023-05-29
Optimize workflows

pvk-developer:
Optimize workflows to run only once when pushing to a pull request.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -1,8 +1,9 @@
 name: Integration Tests
 
 on:
-  - push
-  - pull_request
+  push:
+  pull_request:
+    types: [opened, reopened]
 
 jobs:
   unit:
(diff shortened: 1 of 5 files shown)


frances-h:
While we're on the subject, do we want to keep running all tests whenever any branch is pushed or just when branches tied to PRs are pushed?

amontanez24:
> While we're on the subject, do we want to keep running all tests whenever any branch is pushed or just when branches tied to PRs are pushed?

Personally, I sometimes use the branch to run tests on something that I have no intention to make a PR for. I think that we could make changes to only run slower tests on branches with PRs, but I think everything in SDGym is pretty fast. There is also the question of if we need to even run the tests again on PR opening/re-opening since they run with every push anyway.

amontanez24 (approved):
:shipit:

PR #244, opened 2023-06-01
Change `get_trained_synthesizer` to accept metadata as a dict

### LINKED ISSUE #232: benchmark_single_table crashes with metadata dict
### Environment Details

* SDGym version: master
* Python version: Any
* Operating System: Any

### Error Description

When creating a custom synthesizer, the docs mention that you have to input the metadata as a dictionary. But the code is expecting an object. The code below should work, but it errors instead with: 
`AttributeError: 'SingleTableMetadata' object has no attribute 'get'`

### Steps to reproduce

```python

def get_trained_synthesizer(data, metadata):
    metadata_obj = SingleTableMetadata.load_from_dict(metadata)
    model = GaussianCopulaSynthesizer(metadata_obj)
    model.fit(data)
    return model

def sample_from_synthesizer(synthesizer, n_samples):
    return synthesizer.sample(n_samples)

TestSynthesizer = create_single_table_synthesizer(
    display_name='TestSynthesizer',
    get_trained_synthesizer_fn=get_trained_synthesizer,
    sample_from_synthesizer_fn=sample_from_synthesizer
)

benchmark_single_table(
    custom_synthesizers=[TestSynthesizer]
)
```

fealho:
Resolve #232.


sdgym/synthesizers/base.py
```diff
@@ -39,15 +42,17 @@ def get_trained_synthesizer(self, data, metadata):
...
-                The metadata.
+            metadata (dict):
+                The metadata dictionary.
 
         Returns:
             obj:
-                The synthesizer object
+                The synthesizer object.
         """
+        metadata_modality = MultiTableMetadata() if 'tables' in metadata else SingleTableMetadata()
```
amontanez24, line 54: minor: Can we call `metadata_class`?

tests/unit/test_benchmark.py
```diff
@@ -176,3 +178,43 @@ def test_progress_bar_updates(tqdm_mock):
 
     # Assert
     tqdm_mock.assert_called_once_with(ANY, total=1, position=0, leave=True)
+
+
+def test_benchmark_single_table_custom_synthesizer():
```
amontanez24, line 183: this seems more like an integration test to me since it will be downloading the dataset and metadata behind the scenes

amontanez24 (approved):
A couple minor comments but otherwise looks good!

tests/unit/test_benchmark.py
fealho: Yeah, I felt so as well, but I really wanted to get that extra 11% coverage...

amontanez24: Feel free to add more unit tests 😂 . If you do though, we should mock any public or external functions being used

pvk-developer (approved):
LGTM!

PR #245, opened 2023-06-01
Fix `limit_dataset_size` parameter for `benchmark_single_table`

### LINKED ISSUE #231: limit_dataset_size causes sdgym to crash
### Environment Details

* SDGym version: master
* Python version: Any
* Operating System: Any

### Error Description

Setting the `limit_dataset_size` parameter to True causes the following error:
`(KeyError: 'tables')`

### Steps to reproduce

```python
from sdgym import benchmark_single_table

benchmark_single_table(
    synthesizers=[
        'TVAESynthesizer',
    ],
    limit_dataset_size=True,
)
```

fealho:
Resolve #231.

### DIFF
--- a/sdgym/datasets.py
+++ b/sdgym/datasets.py
@@ -62,18 +62,11 @@ def _get_dataset_path(modality, dataset, datasets_path, bucket=None, aws_key=Non
     return dataset_path
 
 
-def _apply_max_columns_to_metadata(metadata, max_columns):
-    tables = metadata['tables']
-    for table in tables.values():
-        fields = table['fields']
-        if len(fields) > max_columns:
-            fields = dict(itertools.islice(fields.items(), max_columns))
-            table['fields'] = fields
-
-        structure = table.get('structure')
-        if structure:
-            structure['structure'] = structure['structure'][:max_columns]
-            structure['states'] = structure['states'][:max_columns]
+def _apply_max_columns_to_metadata(metadata_dict, max_columns):
+    columns = metadata_dict['columns']
+    if len(columns) > max_columns:
+        columns = dict(itertools.islice(columns.items(), max_columns))
+        metadata_dict['columns'] = columns
 
 
 def load_dataset(modality, dataset, datasets_path=None, bucket=None, aws_key=None,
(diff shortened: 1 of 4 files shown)


sdgym/datasets.py
```diff
@@ -62,22 +63,24 @@ def _get_dataset_path(modality, dataset, datasets_path, bucket=None, aws_key=Non
...
-            structure['structure'] = structure['structure'][:max_columns]
-            structure['states'] = structure['states'][:max_columns]
+    data = data.head(max_rows)
+
+    return data, metadata_dict
 
 
 def load_dataset(modality, dataset, datasets_path=None, bucket=None, aws_key=None,
-                 aws_secret=None, max_columns=None):
+                 aws_secret=None, limit_dataset_size=None):
```
R-Palazzo, line 83: Should we set `limit_dataset_size` to `False` by default? We could also add a `validation` that checks that `limit_dataset_size` is a boolean.

sdgym/datasets.py
```diff
@@ -62,22 +63,24 @@ def _get_dataset_path(modality, dataset, datasets_path, bucket=None, aws_key=Non
...
-def _apply_max_columns_to_metadata(metadata, max_columns):
-    tables = metadata['tables']
-    for table in tables.values():
-        fields = table['fields']
-        if len(fields) > max_columns:
-            fields = dict(itertools.islice(fields.items(), max_columns))
-            table['fields'] = fields
+def _get_dataset_subset(data, metadata_dict):
+    if 'tables' in metadata_dict.keys():
+        raise ValueError('max_columns is not supported for multi-table datasets')
```
R-Palazzo, line 68: Should we modify the error message since `max_columns` is no longer a method parameter?

R-Palazzo (commented):
Looks good! Just let 2 suggestions

sdgym/datasets.py
fealho: `load_dataset` is just a helper function, only `benchmark_single_table` has access to it. I think leaving it as None is better, because it makes it clear the parameter doesn't matter by default, only if you pass a value will it do something.

frances-h (approved):

R-Palazzo (approved):

PR #263, opened 2024-02-23
Metadata sdtypes should be used instead of inferred sdtypes

### LINKED ISSUE #248: The `UniformSynthesizer` should follow the sdtypes in metadata (not the data's dtypes)
### Environment Details
* SDGym version: 0.6.0 (latest)

### What is expected
The `UniformSynthesizer` is expected to uniformly (randomly) create data within the observed ranges or categories.

- For `numerical` or `datetime` data, it should learn the min and max values during fit. Then during sample, it can create random, uniform data in the range
- For `categorical` or `boolean` data, it should learn the possible categories during fit. Then during sample, it can randomly select categories with equal probability (i.e. make it uniform)
- For any other sdtype (such as `id`, `pii`, etc.), it can simply use the `RegexGenerator` or `AnonymizedFaker` to generate values from scratch (no learning or uniform sampling expected)

**How does this synthesizer know which type is which?** It should use the provided `metadata` as the ground source of truth. 

### What is actually observed
Rather than using the metadata to understand the sdtypes, the code just allows the RDT to guess based on the dataframe. [See this line](https://github.com/sdv-dev/SDGym/blob/master/sdgym/synthesizers/uniform.py#L13).

The **automatically-detected RDT config is not guaranteed to be correct.** For example:

- The RDT will detect any integers as being numerical, but they may actually be categorical sdtypes or IDs
- The RDT will detect any strings as being categorical, but they may actually be datetimes, PII or ID types

Instead of `detect_initial_config`, the synthesizer _should_ be parsing the metadata and using the sdtype to decide what to do.

lajohn4747:
resolves #248 

No longer default to inferred type. Rely on metadata unless it isn't specified or the sdtype is not supported.

Add a test to confirm it works as intended.

### DIFF
--- a/sdgym/synthesizers/uniform.py
+++ b/sdgym/synthesizers/uniform.py
@@ -16,23 +16,18 @@ class UniformSynthesizer(BaselineSynthesizer):
     def _get_trained_synthesizer(self, real_data, metadata):
         hyper_transformer = HyperTransformer()
         hyper_transformer.detect_initial_config(real_data)
-        metadata_dict = {}
-        if isinstance(metadata, dict):
-            metadata_dict = metadata
-        else:
-            metadata_dict = metadata.to_dict()
-        if 'columns' in metadata_dict:
-            supported_sdtypes = hyper_transformer._get_supported_sdtypes()
-            config = {}
-            for column in metadata_dict['columns']:
-                sdtype = metadata_dict['columns'][column]['sdtype']
-                if sdtype in supported_sdtypes:
-                    config[column] = metadata_dict['columns'][column]['sdtype']
-                else:
-                    LOGGER.info(
-                        f'Column {column} sdtype: {sdtype} is not supported, '
-                        f'defaulting to inferred type.')
-            hyper_transformer.update_sdtypes(config)
+        supported_sdtypes = hyper_transformer._get_supported_sdtypes()
+        config = {}
+        for column_name, column in metadata.columns.items():
+            sdtype = column['sdtype']
+            if sdtype in supported_sdtypes:
+                config[column_name] = sdtype
+            else:
+                LOGGER.info(
+                    f'Column {column} sdtype: {sdtype} is not supported, '
+                    f'defaulting to inferred type.')
+
+        hyper_transformer.update_sdtypes(config)
 
         # This is done to match the behavior of the synthesizer for SDGym <= 0.6.0
         columns_to_remove = [
--- a/sdgym/synthesizers/uniform.py
+++ b/sdgym/synthesizers/uniform.py
@@ -1,9 +1,12 @@
 """UniformSynthesizer module."""
+import logging
 import numpy as np
 import pandas as pd
 from rdt.hyper_transformer import HyperTransformer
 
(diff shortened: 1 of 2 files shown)


sdgym/synthesizers/uniform.py
```diff
@@ -1,17 +1,39 @@
...
     def _get_trained_synthesizer(self, real_data, metadata):
         hyper_transformer = HyperTransformer()
         hyper_transformer.detect_initial_config(real_data)
+        if metadata:
+            metadata_dict = {}
+            if isinstance(metadata, dict):
+                metadata_dict = metadata
+            else:
+                metadata_dict = metadata.to_dict()
+            if 'columns' in metadata_dict:
```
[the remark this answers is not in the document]
lajohn4747: It seems I see most cases are that metadata is passed into this function as a dict. If a metadata dict is passed in, should we convert it to a SingleTableMetadata/MultiTableMetadata object instead?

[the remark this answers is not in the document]
frances-h: It looks like we already do that in the public method `BaselineSynthesizer.get_trained_synthesizer`, so unless we're calling this method somewhere else it should already be handled.

sdgym/synthesizers/uniform.py
```diff
@@ -1,17 +1,33 @@
...
+        supported_sdtypes = hyper_transformer._get_supported_sdtypes()
+        config = {}
+        for column_name, column in metadata.columns.items():
+            sdtype = column['sdtype']
+            if sdtype in supported_sdtypes:
+                config[column_name] = sdtype
+            else:
+                LOGGER.info(
+                    f'Column {column} sdtype: {sdtype} is not supported, '
+                    f'defaulting to inferred type.')
```
amontanez24, line 28: I think if we do this, it will drop any pii type (eg. phone_number) and use the inferred type instead, which will just be categorical. The issue says this synthesizer can use Faker for those types so I don't think we want to do that

lajohn4747: Updated to use the `PII` type if labeled.

Do you recommend that we infer `PII` if its one of the `SDTYPE_ANONYMIZERS` is detected (e.g. email) but **unlabeled** within the metadata

sdgym/synthesizers/uniform.py
```diff
@@ -1,18 +1,35 @@
...
+        for column_name, column in metadata.columns.items():
+            sdtype = column['sdtype']
+            if sdtype in supported_sdtypes:
+                config[column_name] = sdtype
+            elif column.get('pii', False):
+                config[column_name] = 'pii'
+            else:
+                LOGGER.info(
+                    f'Column {column} sdtype: {sdtype} is not supported, '
+                    f'defaulting to inferred type.')
```
amontanez24, line 30: sorry I meant that even if the sdtype is unsupported, RDT should still work on it, it will just use a faker. So I don't think we need to worry about unsupported types

amontanez24 (approved):
LGTM!

frances-h (approved):

Issue #267, opened 2024-02-26
Remove bumpversion and use bump-my-version


amontanez24:
### Problem Description

- The bumpversion library is not a maintained library and we should switch to using `bump-my-version`.
  - Alternatively, we could also just use Python's native version management (relying on dev's to increment 1 file).

### Expected behavior
- Use `bump-my-version` to version bump sdv.
- Alternatively, just specify the version in 1 location and have `pyproject.toml` read from that location.

PR #269, opened 2024-02-28
Transition from using setup.py to pyproject.toml to specify project metadata

### LINKED ISSUE #266: Transition from using setup.py to pyroject.toml to specify project metadata
### Problem Description
- We currently use **setup.py** and **setup.cfg** to specify the project metadata. This includes requirements which are scatter in `apt.txt`, `requirements.txt`, and `tox.ini`.
- The Python community selected pyproject.toml as a standard way of specifying project metadata, removing the need for all these files. All the project metadata should go in this file. 
- Other modern packaging and dependency management projects, like [Python Poetry](https://python-poetry.org/), only support the more declarative way of using a pyproject.toml file.

### Expected behavior
- Use `pyproject.toml` as the one file for managing project metadata

R-Palazzo:
86azg1hek
Resolve #266

Hi few things to notice/discuss:
- Docker: I changed setup.py to `pyproject.toml`
- `setup.cfg`: I was not able to convert all the config inside `pyproject.toml`, mainly:
    - falke8 config
    - The bumpversion (is it ok because we will change it in [Issue #267](https://github.com/sdv-dev/SDGym/issues/267) no?)
    - `aliases` (don’t no if we’re still using them)

Let me know your thoughts on this, thanks


pyproject.toml
```diff
@@ -0,0 +1,199 @@
+[build-system]
+requires = ['setuptools', 'wheel']
+build-backend = 'setuptools.build_meta'
+
+[project]
+authors = [{ name = 'DataCebo, Inc.', email = 'info@sdv.dev' }]
+classifiers = [
```
gsheni, line 7: keep this at the top
```suggestion
name = 'sdgym'
description = 'Benchmark tabular synthetic data generators using a variety of datasets'
authors = [{ name = 'DataCebo, Inc.', email = 'info@sdv.dev' }]
classifiers = [
```

pyproject.toml
```diff
@@ -0,0 +1,199 @@
...
+]
+description = 'Benchmark tabular synthetic data generators using a variety of datasets'
+keywords = ['machine learning', 'synthetic data generation', 'benchmark', 'generative models']
+name = 'sdgym'
+version = '0.7.1.dev0'
+license = { text = 'BSL-1.1' }
+requires-python = '>=3.8,<3.12'
+readme = 'README.md'
+urls = { 'Homepage' = 'https://github.com/sdv-dev/SDGym' }
+
```
gsheni, line 27:
```suggestion
[project.urls]
"Source Code"= "https://github.com/sdv-dev/SDGym/"
"Issue Tracker" = "https://github.com/sdv-dev/SDGym/issues"
"Twitter" = "https://twitter.com/sdv_dev"
"Chat" = "https://bit.ly/sdv-slack-invite"
```

pyproject.toml
```diff
@@ -0,0 +1,199 @@
...
+    'Topic :: Scientific/Engineering :: Artificial Intelligence',
+]
+description = 'Benchmark tabular synthetic data generators using a variety of datasets'
+keywords = ['machine learning', 'synthetic data generation', 'benchmark', 'generative models']
+name = 'sdgym'
+version = '0.7.1.dev0'
+license = { text = 'BSL-1.1' }
+requires-python = '>=3.8,<3.12'
+readme = 'README.md'
+urls = { 'Homepage' = 'https://github.com/sdv-dev/SDGym' }
```
gsheni, line 26: Remove from here and put below
```suggestion
```

pyproject.toml
```diff
@@ -0,0 +1,199 @@
...
+    'rdt>=1.6.1,<2.0',
+    'sdmetrics>=0.11.0,<1.0',
+    'sdv>=1.3.0,<2',
+]
+
+[project.entry-points]
+sdgym = { main = 'sdgym.cli.__main__:main' }
+
+[project.optional-dependencies]
+
```
gsheni, line 59: I do not think you can have a blank line in TOML like this

tasks.py
```diff
@@ -70,8 +70,8 @@ def _validate_python_version(line):
 
 @task
 def install_minimum(c):
-    with open('setup.py', 'r') as setup_py:
-        lines = setup_py.read().splitlines()
+    with open('pyproject.toml', 'r') as pyproject:
```
gsheni, line 73: Does this code work? For getting the minimum dependencies from `pyproject.toml` you can do something like the following:
```python
import tomli

def parse_pyproject_toml(path):
    requirements = []
    with open(paths, "rb") as f:
        toml_dict = tomli.load(f)

    requirements = toml_dict["project"]["dependencies"]
    optional_reqs = toml_dict["project"]["optional-dependencies"]["all"]
    return requirements, optional_reqs

def clean_list_length_one(item):
    if isinstance(item, list) and len(item) == 1 and " " in item[0]:
        item = item[0].split(" ")
    if item == [""]:
        return None
    return item

requirements = parse_pyproject_toml('pyproject.toml')
```

pyproject.toml
```diff
@@ -0,0 +1,199 @@
...
+all = [
+    # dask
+    'dask',
+    'distributed',
+
+    # test
+    'pytest>=3.4.2',
+    'pytest-cov>=2.6.0',
+    'jupyter>=1.0.0,<2',
+    'rundoc>=0.4.3,<0.5',
```
gsheni, line 135: You can reference previous dependencies as such:
```suggestion
        "sdgym[dask,test]",
```

gsheni (changes requested):

gsheni:
- Regarding the flake8 config, we can switch to ruff in different MR and move the config for that into pyproject.toml
- Regarding bumpversion, yes okay for now. 
- yes, aliases are not used

tasks.py
R-Palazzo: Yes it works. I tried to make the minimum changes here. In commit [5919e05](https://github.com/sdv-dev/SDGym/pull/269/commits/5919e05887c2179fee67aa2e8b5b7d67636675ad) I was doing something similar as your proposal, but this add a dependency (`toml` in my case), so let me know if we can add a dependency.

pyproject.toml
R-Palazzo: Yes, done in [7fac59d](https://github.com/sdv-dev/SDGym/pull/269/commits/7fac59d7ad872f32dc3b7307944cfe1980724f17)

pyproject.toml
R-Palazzo: Yes, done in [7fac59d](https://github.com/sdv-dev/SDGym/pull/269/commits/7fac59d7ad872f32dc3b7307944cfe1980724f17)

pyproject.toml
R-Palazzo: Yes, done in [7fac59d](https://github.com/sdv-dev/SDGym/pull/269/commits/7fac59d7ad872f32dc3b7307944cfe1980724f17)

pyproject.toml
R-Palazzo: Yes, done in [7fac59d](https://github.com/sdv-dev/SDGym/pull/269/commits/7fac59d7ad872f32dc3b7307944cfe1980724f17)

pyproject.toml
R-Palazzo: Yes, done in [7fac59d](https://github.com/sdv-dev/SDGym/pull/269/commits/7fac59d7ad872f32dc3b7307944cfe1980724f17)

tasks.py
gsheni: Yep, my vote would that it is okay to add this dependency (it's only generating minimum dependencies).

R-Palazzo: Alright, done in [3ef342a](https://github.com/sdv-dev/SDGym/pull/269/commits/3ef342a60d307271038e824a25ea77a75876712f)

tasks.py
```diff
@@ -70,30 +71,23 @@ def _validate_python_version(line):
...
+        if _validate_python_version(line):
+            requirement = re.match(r'[^>]*', line).group(0)
+            requirement = re.sub(r"""['",]""", '', requirement)
+            version = re.search(r'>=?(\d\.?)+\w*', line).group(0)
+            if version:
+                version = re.sub(r'>=?', '==', version)
+                version = re.sub(r"""['",]""", '', version)
+                requirement += version
+
+            versions.append(requirement)
```
gsheni, line 90: It might be better to use the Requirement and Specifier class from the packaging library, rather than using regex. The code might be something like this:
```suggestion
    for line in dependencies:
        min_package = find_min_requirement(line)
        min_requirements.append(str(min_package))
    min_requirements = "\n".join(min_requirements) + "\n"
    
from packaging.requirements import Requirement
from packaging.specifiers import Specifier

def find_min_requirement(requirement):
    if is_requirement_path(requirement):
        # skip requirement paths
        # ex '-r core_requirements.txt'
        return
    requirement = remove_comment(requirement)
    if not verify_python_environment(requirement):
        return
    if ">=" in requirement:
        # minimum version specified (ex - 'package >= 0.0.4')
        package = Requirement(requirement)
        version = find_operator_version(package, ">=")
        minimum = create_strict_min(version)
    elif "==" in requirement:
        # version strictly specified
        package = Requirement(requirement)
        version = find_operator_version(package, "==")
        minimum = create_strict_min(version)
    else:
        # minimum version not specified (ex - 'package < 0.0.4')
        # version not specified (ex - 'package')
        raise ValueError(
            "Operator does not exist or is an invalid operator (such as '<'). Please specify the minimum version ('>=', '==')."
        )
    name = determine_package_name(package)
    min_requirement = Requirement(name + str(minimum))
    return min_requirement
      
def is_requirement_path(requirement):
    if ".txt" in requirement and "-r" in requirement:
        return True
    return False

def remove_comment(requirement):
    if "#" in requirement:
        # remove everything after comment character
        requirement = requirement.split("#")[0]
    return requirement
    
def determine_package_name(package):
    name = package.name
    if len(package.extras) > 0:
        name = package.name + "[" + package.extras.pop() + "]"
    return name
    
def verify_python_environment(requirement):
    package = Requirement(requirement)
    if not package.marker:
        # no python version specified in requirement
        return True
    elif package.marker and package.marker.evaluate():
        # evaluate --> evaluating the given marker against the current Python process environment
        return True
    return False
```

gsheni (commented):
Looks good overall. Just 1 suggestion

tasks.py
amontanez24: I think I agree with trying to remove regex if possible. One thing I'm wondering that doesn't really come into play in SDGym but might in other libraries is that sometimes we point to a branch directly to install from. If so, would the minimum tests crash there because that syntax doesn't use '<'? An example would be something like
```
dependencies = [
  "<packagename>@git+<url-to-repo>#egg=<branch or hash>",
]
```

.github/workflows/readme.yml
```diff
@@ -22,5 +22,6 @@ jobs:
       run: |
           python -m pip install --upgrade pip
           python -m pip install invoke rundoc .
+          python -m pip install toml
```
gsheni, line 25: Wrong library. It is this one
```suggestion
          python -m pip install tomli
```

amontanez24 (approved):
:shipit:

gsheni (approved):

PR #272, opened 2024-03-06
Add ability to run on an ec2 instance

### LINKED ISSUE #265: Add run_on_ec2 flag to benchmark_single_table
### Problem Description

_As a user, sometimes the benchmark function take a while and a lot of compute resources so it would be nice to be able to run it on a separate instance._

### Expected behavior

- Add `run_on_ec2` boolean parameter to [benchmark_single_table](https://github.com/sdv-dev/SDGym/blob/581e1e790f06aeaec8c47fede8dcb070eea21087/sdgym/benchmark.py#L445)
- If it is True
    1. Launch an ec2 instance
    2. Install sdgym on that instance
    3. Run the job on that instance with the rest of the parameters
    4. Store the output in an S3 folder based on the value of `output_filepath`
    
### Technical Details
- `output_filepath` is required to be an s3 bucket if the flag is enabled. We should add a check for that
- We should do this using boto3 directly. They have a function called [run_instances](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/ec2/client/run_instances.html) that can take in a script. In our case, we just want the script to pip install sdgym and then run the cli with the commands provided in the method above

lajohn4747:
resolves #265 

Add the field `run_on_ec2` which spins up an EC2 instance with the environment credentials, runs the latest version of sdgym with the same parameters and saves the output to the s3 bucket. Terminates the Ec2 instance once finished with the process to avoid keeping paid resources alive.

Only issue is that custom synthesizers may not work with this command in this PR (as that requires a standardized way to bring in custom imported classes to the EC2 instance as well)

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -10,6 +10,7 @@ import warnings
 from datetime import datetime
 from pathlib import Path
 
+import boto3
 import compress_pickle
 import numpy as np
 import pandas as pd
@@ -447,7 +448,7 @@ def benchmark_single_table(synthesizers=DEFAULT_SYNTHESIZERS, custom_synthesizer
                            limit_dataset_size=False, compute_quality_score=True,
                            sdmetrics=DEFAULT_METRICS, timeout=None, output_filepath=None,
                            detailed_results_folder=None, show_progress=False,
-                           multi_processing_config=None):
+                           multi_processing_config=None, run_on_ec2=False):
     """Run the SDGym benchmark on single-table datasets.
 
     Args:
@@ -487,7 +488,9 @@ def benchmark_single_table(synthesizers=DEFAULT_SYNTHESIZERS, custom_synthesizer
             timeout is enforced.
         output_filepath (str or ``None``):
             A file path for where to write the output as a csv file. If ``None``, no output
-            is written.
+            is written. If run_on_ec2 flag, the filepath should be structured as:
+            {s3_bucket_name}/{path_to_file}
+            Please make sure the path exists and permissions are given.
         detailed_results_folder (str or ``None``):
             The folder for where to store the intermediary results. If ``None``, do not store
             the intermediate results anywhere.
@@ -504,6 +507,14 @@ def benchmark_single_table(synthesizers=DEFAULT_SYNTHESIZERS, custom_synthesizer
         pandas.DataFrame:
             A table containing one row per synthesizer + dataset + metric.
     """
+    if run_on_ec2:
+        if output_filepath is not None:
+            script_content = _create_sdgym_script(dict(locals()), output_filepath)
+            _create_instance_on_ec2(script_content)
+        else:
+            raise ValueError('In order to run on EC2, please provide an S3 folder output.')
+        return None
+
     _validate_inputs(output_filepath, detailed_results_folder, synthesizers, custom_synthesizers)
 
     _create_detailed_results_directory(detailed_results_folder)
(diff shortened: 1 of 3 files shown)


sdgym/benchmark.py
```diff
@@ -523,3 +534,122 @@ def benchmark_single_table(synthesizers=DEFAULT_SYNTHESIZERS, custom_synthesizer
...
+    aws ec2 terminate-instances --instance-ids $INSTANCE_ID
+    """
+
+    response = ec2_client.run_instances(
+        ImageId='ami-07d9b9ddc6cd8dd30',
+        InstanceType='t2.medium',
+        SecurityGroupIds=['sg-097d14e3efa8effb4'],
+        MinCount=1,
+        MaxCount=1,
+        KeyName='john',
```
amontanez24, line 627: we need to replace this right?

sdgym/benchmark.py
```diff
@@ -523,3 +534,122 @@ def benchmark_single_table(synthesizers=DEFAULT_SYNTHESIZERS, custom_synthesizer
         write_csv(scores, output_filepath, None, None)
 
     return scores
+
+
+def _create_sdgym_script(params, output_filepath):
```
amontanez24, line 539: minor: for both of these helpers, can we move them above `benchmark_single_table`?

sdgym/benchmark.py
```diff
@@ -504,6 +507,14 @@ def benchmark_single_table(synthesizers=DEFAULT_SYNTHESIZERS, custom_synthesizer
         pandas.DataFrame:
             A table containing one row per synthesizer + dataset + metric.
     """
+    if run_on_ec2:
+        if output_filepath is not None:
+            script_content = _create_sdgym_script(dict(locals()), output_filepath)
+            _create_instance_on_ec2(script_content)
```
amontanez24, line 513: can we add the following unit tests:
1. Make sure it only runs this when the flag is set
2. Test the value error below
3. test the output of the `_create_sdgym_script`

sdgym/benchmark.py
```diff
@@ -487,7 +488,9 @@ def benchmark_single_table(synthesizers=DEFAULT_SYNTHESIZERS, custom_synthesizer
             timeout is enforced.
         output_filepath (str or ``None``):
             A file path for where to write the output as a csv file. If ``None``, no output
-            is written.
+            is written. If run_on_ec2 flag, the filepath should be structured as:
```
amontanez24, line 491: minor: can we add that `output_filepath` must be provided

sdgym/benchmark.py
```diff
@@ -523,3 +534,122 @@ def benchmark_single_table(synthesizers=DEFAULT_SYNTHESIZERS, custom_synthesizer
...
+    aws configure set aws_secret_access_key {credentials.secret_key}
+    aws configure set region {session.region_name}
+    INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
+    aws ec2 terminate-instances --instance-ids $INSTANCE_ID
+    """
+
+    response = ec2_client.run_instances(
+        ImageId='ami-07d9b9ddc6cd8dd30',
+        InstanceType='t2.medium',
+        SecurityGroupIds=['sg-097d14e3efa8effb4'],
```
amontanez24, line 624: Is this a default security group or one of ours?

amontanez24 (commented):
Looking good! Just minor comments/questions

sdgym/benchmark.py
lajohn4747: Will remove, not needed to create the instance

sdgym/benchmark.py
```diff
@@ -504,6 +507,14 @@ def benchmark_single_table(synthesizers=DEFAULT_SYNTHESIZERS, custom_synthesizer
         pandas.DataFrame:
             A table containing one row per synthesizer + dataset + metric.
     """
+    if run_on_ec2:
+        if output_filepath is not None:
```
amontanez24, line 511: we should also probably print something to the user saying we are going to launch an instance on their account

sdgym/benchmark.py
```diff
@@ -504,6 +624,15 @@ def benchmark_single_table(synthesizers=DEFAULT_SYNTHESIZERS, custom_synthesizer
         pandas.DataFrame:
             A table containing one row per synthesizer + dataset + metric.
     """
+    if run_on_ec2:
+        LOGGER.info("This will create an instance for the current AWS user's account.")
```
amontanez24, line 628: In this case can we just straight up print the message so the user can't turn it off?

lajohn4747: Had to add the `noqa` comment to avoid linter picking it up

sdgym/benchmark.py
```diff
@@ -442,12 +443,129 @@ def _get_empty_dataframe(compute_quality_score, sdmetrics):
...
+    echo "======== Write Script ==========="
+    sudo touch ~/sdgym_script.py
+    echo "{script_content}" > ~/sdgym_script.py
+    echo "======== Run Script ==========="
+    sudo python3 ~/sdgym_script.py
+    echo "======== Terminate ==========="
+    sudo apt install awscli -y
+    aws configure set aws_access_key_id {credentials.access_key}
+    aws configure set aws_secret_access_key {credentials.secret_key}
+    aws configure set region {session.region_name}
```
amontanez24, line 523: does this default to us-east-1?

amontanez24: We might want to include whatever region it is in the message we print to the user so they know where to find the machine

lajohn4747: It defaults to the credentials region of access set as the default for the profile
It can be set: `export AWS_DEFAULT_REGION`

sdgym/benchmark.py
```diff
@@ -442,12 +443,130 @@ def _get_empty_dataframe(compute_quality_score, sdmetrics):
...
+    first_slash_index = output_filepath.find('/')  # Find the index of the first slash
+    bucket_name = ''
+    key_name = ''
+    if first_slash_index != -1:  # Check if a slash is found
+        bucket_name = output_filepath[:first_slash_index]  # Extract directory part
+        key_name = output_filepath[first_slash_index + 1:]  # Extract filename part
+    else:
+        raise ValueError("""Invalid output_filepath.
+                   The path should be structured as: <s3_bucket_name>/<path_to_file>
+                   Please make sure the path exists and permissions are given.""")
```
gsheni, line 456: Can this be put into a utility function and changed to the following? It is currently a bit hard to understand what is going on, and I would suggest something like this: 
```python
    from botocore.exceptions import ClientError
    import boto3

    def parse_s3_path(s3_path):
        if '/' not in s3_path:
            raise ValueError("Invalid S3 path format. Expected 'bucket_name/path_to_file'.")
        bucket_name, key_name = s3_path.split('/', 1)  # Split only on the first '/'
        return bucket_name, key_name  

    def directory_exists(bucket_name, directory_prefix):
        s3_client = boto3.client('s3')
        response = s3.list_objects_v2(Bucket=bucket_name, Prefix=directory_prefix, Delimiter='/')
        return 'Contents' in response or 'CommonPrefixes' in response

    def check_read_permissions(bucket_name):
        s3 = boto3.client('s3')
        try:
            s3.list_objects_v2(Bucket=bucket_name, MaxKeys=1)
            return True  # No exception means we have read permission
        except ClientError as e:
            if e.response['Error']['Code'] in ('AccessDenied', 'AllAccessDisabled'):
                return False  # Access denied or bucket does not exist
            raise  # Re-raise exception if it's something else we didn't expect
```
- You might be able to re-use the boto3.client instead of re-creating it every time like I did.

sdgym/benchmark.py
```diff
@@ -442,12 +443,130 @@ def _get_empty_dataframe(compute_quality_score, sdmetrics):
...
+    aws configure set aws_access_key_id {credentials.access_key}
+    aws configure set aws_secret_access_key {credentials.secret_key}
+    aws configure set region {session.region_name}
+    INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
+    aws ec2 terminate-instances --instance-ids $INSTANCE_ID
+    """
+
+    response = ec2_client.run_instances(
+        ImageId='ami-07d9b9ddc6cd8dd30',
+        InstanceType='t2.medium',
```
gsheni, line 531: Should these be configurable arguments on the function?

sdgym/benchmark.py
```diff
@@ -442,12 +443,130 @@ def _get_empty_dataframe(compute_quality_score, sdmetrics):
...
+                        'Value': 'SDGym_Temp'
+                    }
+                ]
+            }
+        ],
+        BlockDeviceMappings=[
+            {
+                'DeviceName': '/dev/sda1',
+                'Ebs': {
+                    'VolumeSize': 16,  # Specify the desired size in GB
```
gsheni, line 550: Should this be configurable argument on the function?

sdgym/benchmark.py
lajohn4747: This was one I was wondering about. This requires, the user to put in a lot more information about the configuration. The block mapping in line 548, those fields depend on the image. Instance type depends on the image. I think it's fine to hard code because these configurations exist and would require different scripts for each type of VM and errors on the ec2 instance (not enough memory, incorrect os script, etc.) are pretty much on the user to debug.

amontanez24: Let's leave out configuration for now and add it later if we need to

gsheni (approved):

amontanez24 (approved):
🚢

PR #275, opened 2024-03-15
Change EC2 Type for run_ec2 that supports a full run

lajohn4747:
Issue with Ec2 running out of resources for the full SDGYM job. This PR uses a larger instance with more memory to support the full SDGYM run.

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -544,8 +544,8 @@ def _create_instance_on_ec2(script_content):
     """
 
     response = ec2_client.run_instances(
-        ImageId='ami-07d9b9ddc6cd8dd30',
-        InstanceType='t2.xlarge',
+        ImageId='ami-080e1f13689e07408',
+        InstanceType='p2.xlarge',
         MinCount=1,
         MaxCount=1,
         UserData=user_data_script,
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -538,14 +538,14 @@ def _create_instance_on_ec2(script_content):
     echo "{script_content}" > ~/sdgym_script.py
     echo "======== Run Script ==========="
     sudo python3 ~/sdgym_script.py
-    echo "======== Terminate ==========="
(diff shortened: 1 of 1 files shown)


sdgym/benchmark.py
```diff
@@ -538,14 +538,14 @@ def _create_instance_on_ec2(script_content):
...
-    echo "======== Terminate ==========="
+    echo "======== Complete ==========="
     INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
     aws ec2 terminate-instances --instance-ids $INSTANCE_ID
     """
 
     response = ec2_client.run_instances(
         ImageId='ami-07d9b9ddc6cd8dd30',
-        InstanceType='t2.medium',
+        InstanceType='t2.xlarge',
```
kveerama, line 548: @lajohn4747 I think this instance type would not work. From what I can gather this is an instance that does not have GPUs and is from this table shown here: https://aws.amazon.com/fr/ec2/instance-types/t2/  if that is the case, it may not work still. 

I think the goal is to pick larger instance with GPUs from here: https://aws.amazon.com/fr/ec2/instance-types/p2/
I think in the past we used ``p2.xlarge`` . I'd ask @npatki about this.  

One another question I have is we are using an ``ami``. Would it be possible to do a fresh install of SDGym within the launched instance. An AMI would need to be maintained.

lajohn4747: Sounds good! I agree it might take too long based on Felipe's findings. Will update and spin up a run with `p2.xlarge` to take advantage of the larger processor and the GPU.

The `ami` is just an ID to specify the default image for the virtual machine. This command does install a fresh SDGym from the mainline onto the machine so we aren't maintaining anything outside of SDgym

lajohn4747:
It is confirmed to be completed with the default synthesizers and datasets with a p2.xlarge ec2 instance

amontanez24 (approved):

PR #303, opened 2024-05-28
Add ruff for linting, remove flake8, remove isort, remove pylint

gsheni:
CU-86azg1jc5

- Closes https://github.com/sdv-dev/SDGym/issues/268
- Added [ruff](https://github.com/charliermarsh/ruff) for running linting
- Added **ruff** commands to Makefile
- Removed **flake8** as dev dependency, since **ruff** does this, but faster
- Remove **isort** as dev dependency, since **ruff** has the [same function](https://github.com/charliermarsh/ruff#how-does-ruffs-import-sorting-compare-to-isort), but faster
- Removed **pylint** as dev dependency, since **ruff** has [more implemented rules](https://docs.astral.sh/ruff/faq/#how-does-ruffs-linter-compare-to-pylint)
- Removed setup.cfg


sdv-team:
Task linked: [CU-86azg1jc5 SDGym - Switch to using ruff for Python linting and code formatting #268](https://app.clickup.com/t/86azg1jc5)

amontanez24 (commented):
Why does the PR description say it closes this issue: https://github.com/sdv-dev/SDMetrics/issues/268

gsheni:
@amontanez24 fixed

pyproject.toml
```diff
@@ -203,3 +168,53 @@ replace = "__version__ = '{new_version}'"
 [build-system]
 requires = ['setuptools', 'wheel']
 build-backend = 'setuptools.build_meta'
+
+[tool.ruff]
+preview = true
+line-length = 100
+indent-width = 4
+src = ["sdgym"]
+target-version = "py312"
```
amontanez24, line 177: The [documentation says](https://docs.astral.sh/ruff/settings/#target-version) that this should be the minimum version. Should we change to 3.8?

gsheni: Good point. Given that ruff says to use `requires-python`, I just ended up removing it: https://github.com/sdv-dev/SDGym/pull/303/commits/bbdf1388fda6714bf1da48fbec664b78d50e0e4a
> If you're already using a pyproject.toml file, we recommend project.requires-python instead, as it's based on Python packaging standards, and will be respected by other tools. For example, Ruff treats the following as identical to target-version = "py38":

amontanez24 (approved):

PR #337, opened 2024-08-20
Fix `benchmark_single_table` with custom synthesizers and timeout

### LINKED ISSUE #335: `AttributeError` when running custom synthesizer with timeout
### Environment Details

* SDGym version:
* Python version:
* Operating System:

### Error Description
When running a custom synthesizer through the benchmark with timeouts enabled, the custom synthesizer consistently errors with:
```AttributeError: Can't pickle local object 'create_single_table_synthesizer.<locals>.NewSynthesizer'```

Pickle can't pickle the class because it's created in the local scope of the `create_single_table_synthesizer` function. A way around this is to manually define the base class separately, although it doesn't seem to work in iPython environments. 

### Steps to reproduce
```python
from sdgym import benchmark_single_table, create_single_table_synthesizer
from sdv.single_table import GaussianCopulaSynthesizer
from sdv.metadata import SingleTableMetadata


def get_trained_synth(data, metadata):
    metadata = SingleTableMetadata.load_from_dict(metadata)
    synthesizer = GaussianCopulaSynthesizer(metadata)
    synthesizer.fit(data)
    return synthesizer


def sample_synth(trained_synthesizer, num_samples):
    return trained_synthesizer.sample(num_samples)


custom_synthesizer = create_single_table_synthesizer(
  'SimpleGaussianCopula', get_trained_synth, sample_synth
)

benchmark_single_table(
  synthesizers=[],
  sdv_datasets=['fake_hotel_guests'],
  timeout=120,
  sdmetrics=[],
  custom_synthesizers=[custom_synthesizer]
)
```

fealho:
CU-86b1pnw0h, Resolve #335.


sdv-team:
Task linked: [CU-86b1pnw0h SDGym - `AttributeError` when running custom synthesizer with timeout #335](https://app.clickup.com/t/86b1pnw0h)

sdgym/benchmark.py
```diff
@@ -44,6 +45,10 @@
     used_memory,
 )
 
+multiprocessing.set_start_method('spawn', force=True)
+multiprocessing.reduction.ForkingPickler.dumps = cloudpickle.dumps
```
amontanez24, line 49: Is it risky to replace the `dumps` function? It seems like ForkingPickler's loads is just pickle's load so it's probably safe to replace that, but it has a custom dumps function
https://github.com/python/cpython/blob/ec89620e5e147ba028a46dd695ef073a72000b84/Lib/multiprocessing/reduction.py#L49-L54

fealho: I don't think it's risky, their [docstring](https://github.com/cloudpipe/cloudpickle/blob/bc677daaf3018c96671d704a081ff0591eb0855e/cloudpickle/cloudpickle.py#L17) mentions that "It does not include an unpickler, as standard Python unpickling suffices.
". That means their pickling is pretty standard, so it's unlikely to cause issues.

I added a context manager just in case though, so the changes are localized. Let me know what you think.

amontanez24: Out of curiosity, does it work if you only change the loads function but keep their dumps function? If not I think this is good

fealho: No. It works the other way around, if you only change the `dumps` function, but I decided to change `loads` as well to be consistent.

amontanez24 (approved):

frances-h (approved):

PR #374, opened 2025-01-21
Make the `ColumnSynthesizer` follow the sdtypes in the metadata (not the data's dtypes)

### LINKED ISSUE #249: The `ColumnSynthesizer` should follow the sdtypes in the metadata (not the data's dtypes)
### Environment Details

* SDGym version: 0.6.0 (latest)

### What is expected
The `ColumnSynthesizer` is expected to independently model each column.

- For `numerical` or `datetime` sdtypes, it should learn a univariate GMM during fit. Then during sample, it can create data from it.
- For `categorical` or `boolean` sdtypes, it should learn the frequencies of each category. Then during sample, it can create data using those frequencies as weights.  
- For other sdtypes (such as `id`, `pii`, etc.), it can simply use the `RegexGenerator` or `AnonymizedFaker` to generate values from scratch (no learning is expected)

**How does this synthesizer know which type is which?** It should use the provided metadata as the ground source of truth.

### What is actually observed
Similar to the `UniformSynthesizer` (see #248), this synthesizer just lets the RDT HyperTransformer decide which column is which sdtype (based on the data). 

It should be referencing the metadata, since the metadata is the source of truth.

fealho:
CU-85zt7j8em, Resolve #249.

### DIFF
--- a/sdgym/synthesizers/column.py
+++ b/sdgym/synthesizers/column.py
@@ -1,10 +1,14 @@
 """ColumnSynthesizer module."""
 
+import logging
+
 import pandas as pd
 from rdt.hyper_transformer import HyperTransformer
 from sklearn.mixture import GaussianMixture
 
 from sdgym.synthesizers.base import BaselineSynthesizer
+
+LOGGER = logging.getLogger(__name__)
 
 
 class ColumnSynthesizer(BaselineSynthesizer):
@@ -17,6 +21,21 @@ class ColumnSynthesizer(BaselineSynthesizer):
     def _get_trained_synthesizer(self, real_data, metadata):
         hyper_transformer = HyperTransformer()
         hyper_transformer.detect_initial_config(real_data)
+        supported_sdtypes = hyper_transformer._get_supported_sdtypes()
+        config = {}
+        for column_name, column in metadata.columns.items():
+            sdtype = column['sdtype']
+            if sdtype in supported_sdtypes:
+                config[column_name] = sdtype
+            elif column.get('pii', False):
+                config[column_name] = 'pii'
+            else:
+                LOGGER.info(
+                    f'Column {column} sdtype: {sdtype} is not supported, '
+                    f'defaulting to inferred type.'
+                )
+
+        hyper_transformer.update_sdtypes(config)
 
         # This is done to match the behavior of the synthesizer for SDGym <= 0.6.0
         columns_to_remove = [
--- a/sdgym/synthesizers/column.py
+++ b/sdgym/synthesizers/column.py
@@ -4,6 +4,7 @@ import logging
 
 import pandas as pd
 from rdt.hyper_transformer import HyperTransformer
+from sdv.metadata import Metadata
 from sklearn.mixture import GaussianMixture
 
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-85zt7j8em SDGym - The `ColumnSynthesizer` should follow the sdtypes in the metadata (not the data's dtypes) #249](https://app.clickup.com/t/85zt7j8em)

sdgym/synthesizers/column.py
```diff
@@ -17,6 +21,21 @@ class ColumnSynthesizer(BaselineSynthesizer):
...
+            sdtype = column['sdtype']
+            if sdtype in supported_sdtypes:
+                config[column_name] = sdtype
+            elif column.get('pii', False):
+                config[column_name] = 'pii'
+            else:
+                LOGGER.info(
+                    f'Column {column} sdtype: {sdtype} is not supported, '
+                    f'defaulting to inferred type.'
+                )
```
lajohn4747, line 36: Should we do something here besides log like assign a default config?

lajohn4747 (changes requested):
Have one comment about this PR, also are there failing test checks a problem?

sdgym/synthesizers/column.py
fealho: The inferred sdtype based on the data was [already detected](https://github.com/sdv-dev/SDGym/pull/374/files#diff-70d4f5be152c52d549f3482715ef5631e5231ccb388664c6095fab83a363aadcR23). This is similarily done for the `UniformSynthesizer` [here](https://github.com/sdv-dev/SDGym/blob/241719b82807099b5b1ee090a8cad0ed6b5e6268/sdgym/synthesizers/uniform.py#L28-L32).

The failing problems on main are unrelated.

lajohn4747 (approved):

sdgym/synthesizers/column.py
```diff
@@ -17,6 +21,21 @@ class ColumnSynthesizer(BaselineSynthesizer):
     def _get_trained_synthesizer(self, real_data, metadata):
         hyper_transformer = HyperTransformer()
         hyper_transformer.detect_initial_config(real_data)
+        supported_sdtypes = hyper_transformer._get_supported_sdtypes()
+        config = {}
+        for column_name, column in metadata.columns.items():
```
R-Palazzo, line 26: Is this going to work with the new `Metadata` object?
You might need to provide a table name.

tests/unit/synthesizers/test_column.py
```diff
@@ -13,9 +14,11 @@ def test__get_trained_synthesizer(self, gm_mock):
         column_synthesizer = ColumnSynthesizer()
         column_synthesizer.length = 10
         data = pd.DataFrame({'col1': [1, 2, 3, 4]})
+        metadata = SingleTableMetadata()
```
R-Palazzo, line 17: Can we use the new `Metadata` object here?

R-Palazzo (commented):
Looks good!
Just let a comment about handling the new `Metadata` object

R-Palazzo (approved):

PR #389, opened 2025-04-04
Update the library installation script for EC2 machines to install optional dependencies like RealTabFormer

### LINKED ISSUE #388: Update the library installation script for EC2 machines to install optional dependencies like RealTabFormer
### Problem Description

In our goal of including more synthesizers and datasets in our sdgym monthly run, we should ensure that any new synthesizer can be run on EC2 machines. Currently, this is not the case since new synthesizers are optional dependencies of the library that we do not install on the EC2 machines.

### Expected behavior
Update the installation script on EC2 machines to check for any optional dependencies needed to run the external synthesizers and install the appropriate sdgym accordingly.

### Additional context
Below is the code that we run on the ec2 instance. We can update it to install the optional dependencies as well.
https://github.com/sdv-dev/SDGym/blob/40327574f687d80ddbf5b85998c395d7f41f7a5d/sdgym/benchmark.py#L674-L692
We could either always install the optional dependencies, or pass an argument with the synthesizers that require optional dependencies to be installed, and only install them if the argument is provided.

R-Palazzo:
Resolve #388
Did one successfull run with 1 dataset, the results are stored [here](https://us-east-1.console.aws.amazon.com/s3/object/sdgym-benchmark?region=us-east-1&bucketType=general&prefix=Debug/Try_2_RealTabFormerSynthesizer_dataset_expedia_hotel_logs.csv)
s3://sdgym-benchmark/Debug/Try_2_RealTabFormerSynthesizer_dataset_expedia_hotel_logs.csv

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -647,6 +647,7 @@ import sdgym
 from sdgym.synthesizers.sdv import (CopulaGANSynthesizer, CTGANSynthesizer,
     GaussianCopulaSynthesizer, HMASynthesizer, PARSynthesizer, SDVRelationalSynthesizer,
     SDVTabularSynthesizer, TVAESynthesizer)
+from sdgym.synthesizers import RealTabFormerSynthesizer
 
 results = sdgym.benchmark_single_table(
     {synthesizer_string}, custom_synthesizers={params['custom_synthesizers']},
@@ -675,7 +676,7 @@ def _create_instance_on_ec2(script_content):
     sudo apt update -y
     sudo apt install python3-pip -y
     echo "======== Install Dependencies ============"
-    sudo pip3 install sdgym
+    sudo pip3 install sdgym[all]
     sudo pip3 install anyio
     pip3 list
     sudo apt install awscli -y
@@ -705,7 +706,7 @@ def _create_instance_on_ec2(script_content):
             {
                 'DeviceName': '/dev/sda1',
                 'Ebs': {
-                    'VolumeSize': 16,  # Specify the desired size in GB
+                    'VolumeSize': 32,  # Specify the desired size in GB
                     'VolumeType': 'gp2',  # Change the volume type as needed
                 },
             }


pvk-developer:
> Resolve #388 Did one successfull run with 1 dataset, the results are stored [here](https://us-east-1.console.aws.amazon.com/s3/object/sdgym-benchmark?region=us-east-1&bucketType=general&prefix=Debug/Try_2_RealTabFormerSynthesizer_dataset_expedia_hotel_logs.csv) s3://sdgym-benchmark/Debug/Try_2_RealTabFormerSynthesizer_dataset_expedia_hotel_logs.csv

I think that the reason why we are seeing such a long time is mainly this line:
https://github.com/sdv-dev/SDGym/blob/40327574f687d80ddbf5b85998c395d7f41f7a5d/sdgym/synthesizers/realtabformer.py#L40 
As you can see it forces `realtabformer` to run on a `cpu` rather than the `gpu` which should be used by default on the `p2.large` instance. I think that our ctgan model does that automatically that if. You have usable gpu it will use it if not it will go for cpu. Can we try to re-run this with `gpu` instead of forcing it to be `cpu` ?

sdgym/benchmark.py
```diff
@@ -705,7 +706,7 @@ def _create_instance_on_ec2(script_content):
             {
                 'DeviceName': '/dev/sda1',
                 'Ebs': {
-                    'VolumeSize': 16,  # Specify the desired size in GB
+                    'VolumeSize': 32,  # Specify the desired size in GB
```
amontanez24, line 709: is this still necessary with the new instance?

amontanez24: @R-Palazzo any word on if this can be removed

sdgym/benchmark.py
```diff
@@ -724,23 +725,28 @@ def _create_instance_on_ec2(script_content):
...
-    sudo pip3 install anyio
-    pip3 list
-    sudo apt install awscli -y
+    sudo apt install -y python3-pip python3-venv awscli
+    echo "======== Create Virtual Environment ============"
+    python3 -m venv ~/env
+    source ~/env/bin/activate
+    echo "======== Install Dependencies in venv ============"
+    pip install --upgrade pip
+    pip install "git+https://github.com/sdv-dev/SDGym.git@main#egg=sdgym[all]"
```
amontanez24, line 739: you can change this line back before merging

PR #398, opened 2025-05-15
Update the EC2 instance used when run_on_ec2 is enabled

### LINKED ISSUE #396: Update the EC2 instance used when run_on_ec2 is enabled
### Environment Details

* SDGym version: 0.10.0

### Error Description

If you run the [benchmark_single_table](https://github.com/sdv-dev/SDGym/blob/26973b89d3d58dbdc9c93db329283d30ab9b0123/sdgym/benchmark.py#L722) function with the `run_on_ec2` flag enabled, you will see the following error:

ClientError: An error occurred (Unsupported) when calling the RunInstances operation: The requested configuration is currently not supported. Please check the documentation for supported configurations.

This seems to be because AWS removed support for the p2.xlarge instance [from most regions](https://docs.aws.amazon.com/ec2/latest/instancetypes/ec2-instance-regions.html). We can see on an [old version of that same page](https://web.archive.org/web/20250221180458/https://docs.aws.amazon.com/ec2/latest/instancetypes/ec2-instance-regions.html), that the P2 instances were more widely supported. This is the instance sdgym uses
https://github.com/sdv-dev/SDGym/blob/26973b89d3d58dbdc9c93db329283d30ab9b0123/sdgym/benchmark.py#L695-L698

We should update the instance we use to be one of the current generation instance types that is similar in specs and cost. A solid choice is `g4dn.4xlarge`

| Instance Type | Hourly Cost | # GPUs | vCPU | RAM | 
---------------|--------------|---------|-------|-----|
| p2.xlarge | $0.90 | 1 | 4 | 61 GiB |
| g4dn.xlarge | $0.526 | 1 | 4 | 16 GiB |
| g4dn.2xlarge | $0.752 | 1 | 8 | 32 GiB |
| g4dn.4xlarge | $1.204 | 1 | 16 | 64 GiB |

Any of the `g4dn` options are good, but we want to keep similar RAM since that is often a limiter when running synthesizers.

### Steps to reproduce

```python
import sdgym

sdgym.benchmark_single_table(
    run_on_ec2=True
)
```

R-Palazzo:
Resolve #396
CU-86b4xhaae
Launch a successful run with GaussianCopula

Screenshot 2025-05-15 at 17 00 59

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -694,7 +694,7 @@ def _create_instance_on_ec2(script_content):
 
     response = ec2_client.run_instances(
         ImageId='ami-080e1f13689e07408',
-        InstanceType='p2.xlarge',
+        InstanceType='g4dn.4xlarge',
         MinCount=1,
         MaxCount=1,
         UserData=user_data_script,


amontanez24 (commented):
Can we post the comparison between this run and the last run here?

fealho (approved):

R-Palazzo:
@amontanez24 
Screenshot 2025-05-15 at 18 48 56

amontanez24 (approved):

PR #402, opened 2025-05-27
Add parameter to compute privacy score with DCRBaselineProtection metric

### LINKED ISSUE #397: Add DCRBaseline Metric to single table report
### Problem Description

_As a user of synthetic data, I'd like to get an idea of how different synthesizers compare when it comes to data privacy._

We want to add the DCRBaselineProtection metric for privacy to the single table SDGym runs.

### Expected behavior

- Add a new parameter called `compute_privacy_score` that is `True` by default (but can be turned off).
    - Then this parameter is True, there should be a column for the `DCRBaselineProtection` in the report
    - Figure out what parameters to use that are good for performance (total time)
- Deprecate the `sdmetrics` parameter in `benchmark_single_table`.
    - Remove [`NewRowSynthesis`](https://github.com/sdv-dev/SDGym/blob/26973b89d3d58dbdc9c93db329283d30ab9b0123/sdgym/benchmark.py#L62C22-L62C37) from the defaults
    - Say it will be removed in version 0.12.0
    https://github.com/sdv-dev/SDGym/blob/26973b89d3d58dbdc9c93db329283d30ab9b0123/sdgym/benchmark.py#L730
- Add a parameter called `additional_sdmetrics`
    - These will be the extra metrics that the user chooses to run.
    - If someone uses the old `sdmetrics` parameter, raise a FutureWarning and use the list they pass here instead

gsheni:
- Fixes #397 
- CU-86b4ya3e6
- Was able to successfully run all datasets on EC2 (with `limit_dataset_size=True`, **GaussianCopulaSynthesizer**): 
  - [Results](https://sdgym-benchmark.s3.us-east-1.amazonaws.com/Debug/privacy_score_may_29.csv)
  - s3://sdgym-benchmark/Debug/privacy_score_may_29.csv


sdgym/benchmark.py
```diff
@@ -764,13 +816,19 @@ def benchmark_single_table(
...
+        compute_privacy_score (bool):
+            Whether or not to evaluate an overall privacy score. Defaults to ``True``.
+        additional_sdmetrics (list[str]):
+            A list of the additional SDMetrics to use. If you'd like to input specific parameters
             into the metric, provide a tuple with the metric name followed by a dictionary of
             the parameters.
+        sdmetrics (list[str]):
+            **DEPRECATED**  A list of the different SDMetrics to use.
+            If you'd like to input specific parameters into the metric, provide a tuple with
+            the metric name followed by a dictionary of the parameters.
```
amontanez24, line 831: I forgot to update the issue but I don't think we need to deprecate this anymore since we aren't going to run the new metric as part of this list.

static_code_analysis.txt
```diff
@@ -32,7 +32,7 @@ Test results:
 156	        assert issubclass(type(synthesizer), BaselineSynthesizer), (
 157	            '`synthesizer` must be an instance of a synthesizer class.'
 158	        )
-159	
+159
```
amontanez24, line 35: does this file need to be excluded from the linter?

sdgym/benchmark.py
amontanez24: no, you can just leave it as sdmetrics. The only change here will be removing the NewRowSynthesis metrics from the default list and the default list altogether

static_code_analysis.txt
gsheni: Yep, fixed:
https://github.com/sdv-dev/SDGym/pull/402/commits/98dc8bf052b77690b5595381480ada1f7c9f3fe7#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R199

sdgym/benchmark.py
```diff
@@ -220,33 +225,47 @@ def _compute_scores(
...
 
         quality_report.generate(real_data, synthetic_data, metadata, verbose=False)
-        output['quality_score_time'] = (datetime.utcnow() - start).total_seconds()
+        output['quality_score_time'] = calculate_score_time(start)
         output['quality_score'] = quality_report.get_score()
 
+    if compute_privacy_score:
+        start = get_utc_now()
+        num_rows = len(synthetic_data)
+        num_rows_subsample = math.floor(num_rows * 0.60)
```
fealho, line 258: Where is the `0.60` coming from? Maybe add an explanation for it.

fealho (approved):

PR #415, opened 2025-06-27
Add ability to save synthesizers and data when running benchmark_single_table

### LINKED ISSUE #410: Add ability to save synthesizers and data when running benchmark_single_table
### Problem Description

_As a user, I'd like to be able to analyze the synthetic data and reproduce a result for a specific SDGym run._

To do this, we need to start saving the  synthesizers and synthetic data that gets created.

### Expected behavior

- Add parameter `output_destination` to `benchmark_single_table`.
    - If this parameter is set, we should store the synthetic data as a csv and the synthesizers as a .pkl file to that path.
    - We should also save a yaml file with the following. This file should have a unique name for the run like run_{run_id}.yaml:
        - SDV version (if SDV synthesizer otherwise version of corresponding library for example REalTabFormer)
        - SDGym version
        - Date job was completed
        - A list of all the jobs that the run was kicked off with. This is because multiple runs can be kicked off the same day and it would be useful to know when they all finish and what each run was made up of. This file is more for debugging purposes.
    - The path must be a local folder path. If it is an s3 path, crash and say to use the `benchmark_single_table_aws` function instead.
- When saving the synthesizers and synthetic data, we should save with the following folder structure inside of the provided folder. Name refers to the name of the synthesizer (eg. ctgan):
    - `SDGym_results_mm_dd_yyyy/<dataset_name_mm_dd_yyyy>/<synthesizer_name>_synthesizer.pkl`
    - `SDGym_results_mm_dd_yyyy/<dataset_name_mm_dd_yyyy>/<synthesizer_name>_synthetic_data.csv`
    - `SDGym_results_mm_dd_yyyy/<dataset_name_mm_dd_yyyy>/meta.yaml`
- The folder that will contain all of the results (eg. SDGym_results_mm_dd_yyyy) should be created as soon as the run begins.
    - We should update the job completed date in the yaml file once the final job (synthesizer/dataset combination) completes. 
- We should build the results.csv file as each job completes and put that in the SDGym_run_mm_dd_yyyy folder. This is the final csv with all the scores and times.
    - We should be threadsafe here.
- We should also use this time to deprecate the following parameters. For all of them we should say support will be dropped in 3 months.
    - output_filepath
    - detailed_results_folder
    - multi_processing_config
    - run_on_ec2

### Additional context

- We will need to start passing this filepath down to the _synthesize function
https://github.com/sdv-dev/SDGym/blob/e692ad8115b760e3640e6c7429fc53016ca8fcfb/sdgym/benchmark.py#L152
- We can either save the data and synthesizer from there, or move it to the [synthesizer ](https://github.com/sdv-dev/SDGym/blob/e692ad8115b760e3640e6c7429fc53016ca8fcfb/sdgym/synthesizers/base.py#L12)class
- We need to make sure any dependencies we add are also added to the [ec2 script](https://github.com/sdv-dev/SDGym/blob/e692ad8115b760e3640e6c7429fc53016ca8fcfb/sdgym/benchmark.py#L723) or directly to the pyproject.toml
- If there is a collision (the same synthesizer and dataset run on the same date) just overwrite the values

R-Palazzo:
Resolve #410
CU-86b5dy0pa

Thanks in advance for the review. Here are a few questions:
1. Should we crash if the `output_destination` already exists? Or should we always be overwriting (for instance if two benchmark are launched the same day)
2. What should include the `meta.yaml` file that is expected to be saved in `SDGym_results_mm_dd_yyyy/<dataset_name_mm_dd_yyyy>`? I did not create it yet because I was not sure what to put inside it.
3. Compared to the naming given in the issue:
    - I don't save with underscores at the end (`synthetic_data.csv` instead of `_synthetic_data.csv`), is it okay?
    - The `run<id>.yaml` is at the output_destination as well as the `SDGym_results_mm_dd_yyyy` is it correct or should it be inside `SDGym_results_mm_dd_yyyy`?
    - All the results combined are saved in a results.csv file that is in `SDGym_results_mm_dd_yyyy`
4. In `run<id>.yaml` I defined the starting_date and completed_date that correspond to the time the benchmark was started and fully commpleted.

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -856,14 +856,13 @@ def _create_instance_on_ec2(script_content):
 
 
 def _handle_deprecated_parameters(
-    output_filepath, detailed_results_folder, multi_processing_config, run_on_ec2
+    output_filepath, detailed_results_folder, multi_processing_config
 ):
     """Handle deprecated parameters and issue warnings."""
     parameters_to_deprecate = {
         'output_filepath': output_filepath,
         'detailed_results_folder': detailed_results_folder,
         'multi_processing_config': multi_processing_config,
-        'run_on_ec2': run_on_ec2,
     }
     parameters = []
     for name, value in parameters_to_deprecate.items():
@@ -873,9 +872,10 @@ def _handle_deprecated_parameters(
     if parameters:
         parameters = "', '".join(sorted(parameters))
         message = (
-            f"Parameters '{parameters}' are deprecated in the `benchmark_single_table` "
+            f"Parameters '{parameters}' are deprecated in the 'benchmark_single_table' "
             'function and will be removed in October 2025. '
-            'Please consider using `output_destination` instead.'
+            "For saving results, please use the 'output_destination' parameter. For running SDGym"
+            " remotely on AWS please use the 'benchmark_single_table_aws' method."
         )
         warnings.warn(message, FutureWarning)
 
@@ -1024,9 +1024,7 @@ def benchmark_single_table(
         pandas.DataFrame:
             A table containing one row per synthesizer + dataset + metric.
     """
-    _handle_deprecated_parameters(
-        output_filepath, detailed_results_folder, multi_processing_config, run_on_ec2
-    )
+    _handle_deprecated_parameters(output_filepath, detailed_results_folder, multi_processing_config)
     if run_on_ec2:
         print("This will create an instance for the current AWS user's account.")  # noqa
         if output_filepath is not None:
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -892,9 +892,6 @@ def _validate_output_destination(output_destination):
             'Please use `benchmark_single_table_aws` instead.'
         )
 
-    if os.path.exists(output_destination):
-        raise ValueError(f'The output path {output_destination} already exists.')
-
 
 def _write_run_id_file(output_destination, synthesizers, job_args_list):
     jobs = [[job[-3], job[0]['name']] for job in job_args_list]
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -7,10 +7,12 @@ import multiprocessing
 import os
 import pickle
 import tracemalloc
+import uuid
 import warnings
 from collections import defaultdict
 from contextlib import contextmanager
 from datetime import datetime
+from importlib.metadata import version
 from pathlib import Path
 
 import boto3
@@ -19,6 +21,7 @@ import compress_pickle
 import numpy as np
 import pandas as pd
 import tqdm
+import yaml
 from sdmetrics.reports.multi_table import (
     DiagnosticReport as MultiTableDiagnosticReport,
 )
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b5dy0pa SDGym - Add ability to save synthesizers and data when running benchmark_single_table #410](https://app.clickup.com/t/86b5dy0pa)

amontanez24:
> Resolve #410 CU-86b5dy0pa
> 
> Thanks in advance for the review. Here are a few questions:
> 
> 1. Should we crash if the `output_destination` already exists? Or should we always be overwriting (for instance if two benchmark are launched the same day)
> 2. What should include the `meta.yaml` file that is expected to be saved in `SDGym_results_mm_dd_yyyy/<dataset_name_mm_dd_yyyy>`? I did not create it yet because I was not sure what to put inside it.
> 3. Compared to the naming given in the issue:
>    
>    * I don't save with underscores at the end (`synthetic_data.csv` instead of `_synthetic_data.csv`), is it okay?
>    * The `run<id>.yaml` is at the output_destination as well as the `SDGym_results_mm_dd_yyyy` is it correct or should it be inside `SDGym_results_mm_dd_yyyy`?
>    * All the results combined are saved in a results.csv file that is in `SDGym_results_mm_dd_yyyy`
> 4. In `run<id>.yaml` I defined the starting_date and completed_date that correspond to the time the benchmark was started and fully commpleted.

1. If the `output_destination` is created I think we can still write to it. If the same synthesizer/dataset combination is passed then we can overwrite it. Otherwise we can just make the new folders within the `SDGym_results_mm_dd_yyyy`folder
2. This might not have been worded well but I think the run_<id>.yaml can replace the meta.yaml. It should have the sdgym version, the synthesizer library version and the list of jobs (eg. [(ctgan, adult, (ctgan, census)...])
3.
    a. I think the issue maybe got badly formatted. It should be <synthesizer_name>_synthetic_data.csv. For example, ctgan_synthetic_data.csv.
    b. I think it's fine to have both. I'm also ok if you only have the outer one.
    c. That's good
4. This is good.

sdgym/benchmark.py
```diff
@@ -779,6 +855,78 @@ def _create_instance_on_ec2(script_content):
...
+        if value is not None and value:
+            parameters.append(name)
+
+    if parameters:
+        parameters = "', '".join(sorted(parameters))
+        message = (
+            f"Parameters '{parameters}' are deprecated in the `benchmark_single_table` "
+            'function and will be removed in October 2025. '
+            'Please consider using `output_destination` instead.'
+        )
```
R-Palazzo, line 879: Let me know if this warning message makes sense. I introduce `output_destination`, but not all the deprecated parameters relate to saving data.

sdgym/benchmark.py
```diff
@@ -779,6 +855,78 @@ def _create_instance_on_ec2(script_content):
...
+    with open(f'{output_destination}/run_{run_id}.yaml', 'w') as file:
+        yaml.dump(metadata, file)
+
+    return run_id
+
+
+def _update_run_id_file(output_destination, run_id):
+    run_file = Path(output_destination) / f'run_{run_id}.yaml'
+    with open(run_file, 'r') as f:
+        run_data = yaml.safe_load(f) or {}
```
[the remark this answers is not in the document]
R-Palazzo: No we're safe here because the method is called after all the jobs are run and the results generated.

sdgym/benchmark.py
```diff
@@ -691,6 +678,15 @@ def _run_jobs(multi_processing_config, job_args_list, show_progress):
...
     scores = pd.concat(scores, ignore_index=True)
+    output_directions = job_args_list[0][-1]
+    if output_directions and isinstance(output_directions, dict):
+        synth_path = Path(output_directions['synthesizer'])
+        root_path = synth_path.parents[2]
+        result_file = root_path / 'results.csv'
+        if not result_file.exists():
+            scores.to_csv(result_file, index=False, mode='w')
+        else:
+            scores.to_csv(result_file, index=False, mode='a', header=False)
```
amontanez24, line 689: Can we write to a different file here as well to assure no collisions? I was also thinking for the run id that we could just make it the date followed by a number that we increment. Like 20251202_1 or something like that. That might be more readable than a uuid

amontanez24 (approved):

rwedge (approved):

PR #417, opened 2025-07-02
Add benchmark_single_table_aws function

### LINKED ISSUE #414: Add benchmark_single_table_aws function
### Problem Description

__As a user, I'd like to be able to run remote sdgym jobs without misconfiguring anything by accident._

Currently, we have one function ([benchmark_single_table](https://github.com/sdv-dev/SDGym/blob/e692ad8115b760e3640e6c7429fc53016ca8fcfb/sdgym/benchmark.py#L782C5-L782C27)) for all sdgym runs. The issue is that if the run is remote (on ec2), only certain parameters work. The custom_synthesizer will not work and the output_filepath becomes required since there is no session connecting you to the ec2 instance. Providing the AWS credentials also becomes mysterious.

### Expected behavior

Add a whole new function
```python
def benchmark_single_table_aws(
    output_destination='s3://sdgym',
    aws_access_key_id='key_id',
    aws_secret_access_key='secrete_key',
    synthesizers=DEFAULT_SYNTHESIZERS,
    sdv_datasets=DEFAULT_DATASETS,
    additional_datasets_folder=None,
    limit_dataset_size=False,
    compute_quality_score=True,
    compute_diagnostic_score=True,
    compute_privacy_score=True,
    sdmetrics=None,
    timeout=None,
):
    """Run the SDGym benchmark on single-table datasets.

    Args:
        output_destination (str):
            An S3 bucket or filepath. The results output folder will be written here. Should be structured as: 
             s3://{s3_bucket_name}/{path_to_file} or s3://{s3_bucket_name}.
        aws_access_key_id (str): The AWS access key id. Optional
        aws_secret_access_key (str): The AWS secret access key. Optional
        synthesizers (list[string]):
            The synthesizer(s) to evaluate. Defaults to ``[GaussianCopulaSynthesizer,
            CTGANSynthesizer]``. The available options are:
                - ``GaussianCopulaSynthesizer``
                - ``CTGANSynthesizer``
                - ``CopulaGANSynthesizer``
                - ``TVAESynthesizer``
                - ``RealTabFormerSynthesizer``
        sdv_datasets (list[str] or ``None``):
            Names of the SDV demo datasets to use for the benchmark. Defaults to
            ``[adult, alarm, census, child, expedia_hotel_logs, insurance, intrusion, news,
            covtype]``. Use ``None`` to disable using any sdv datasets.
        additional_datasets_folder (str or ``None``):
            The path to an S3 bucket). Datasets found in this folder are
            run in addition to the SDV datasets. If ``None``, no additional datasets are used.
        limit_dataset_size (bool):
            Use this flag to limit the size of the datasets for faster evaluation. If ``True``,
            limit the size of every table to 1,000 rows (randomly sampled) and the first 10
            columns.
        compute_quality_score (bool):
            Whether or not to evaluate an overall quality score. Defaults to ``True``.
        compute_diagnostic_score (bool):
            Whether or not to evaluate an overall diagnostic score. Defaults to ``True``.
        compute_privacy_score (bool):
            Whether or not to evaluate an overall privacy score. Defaults to ``True``.
        sdmetrics (list[str]):
            A list of the different SDMetrics to use.
            If you'd like to input specific parameters into the metric, provide a tuple with
            the metric name followed by a dictionary of the parameters.
        timeout (int or ``None``):
            The maximum number of seconds to wait for synthetic data creation. If ``None``, no
            timeout is enforced.

    Returns:
        pandas.DataFrame:
            A table containing one row per synthesizer + dataset + metric.
    """
```

### Additional context

- This issue should be completed after #410 and should use the same output structure and synthesizer saving process except in an s3 bucket instead of local folder.

R-Palazzo:
Resolve #414
CU-86b5ewpu4

@amontanez24 
- what is the expected behavior for the `additional_datasets_folder` parameter in `benchmark_single_table_aws()`? Should it be an S3 bucket, or could it be a local dataset also?
- Just checking, let me know if the parameter order makes sense, in `benchmark_single_table()` `output_folder` comes after `timeout` while it's the first parameter in `benchmark_single_table_aws()`


sdgym/benchmark.py
```diff
@@ -108,7 +113,32 @@ def _create_detailed_results_directory(detailed_results_folder):
         os.makedirs(detailed_results_folder, exist_ok=True)
 
 
-def _setup_output_destination(output_destination, synthesizers, datasets):
+def _setup_output_destination_aws(output_destination, synthesizers, datasets, s3_client):
+    paths = defaultdict(dict)
+    s3_path = output_destination[5:]
```
amontanez24, line 118: can we make this less mysterious? Maybe make a constant with the prefix and then do `output_destination[len(CONSTANT):]`

sdgym/benchmark.py
```diff
@@ -208,7 +242,54 @@ def _generate_job_args_list(
     return job_args_list
 
 
-def _synthesize(synthesizer_dict, real_data, metadata, synthesizer_path=None):
+def _parse_s3_uri(s3_uri):
+    assert s3_uri.startswith('s3://')
+    parts = s3_uri[5:].split('/', 1)
+    bucket = parts[0]
+    key = parts[1] if len(parts) > 1 else ''
+    return bucket, key
```
amontanez24, line 250: we can maybe move all of this to an aws file

sdgym/benchmark.py
```diff
@@ -630,10 +726,19 @@ def _run_job(args):
...
+            parsed_url = urlparse(synthesizer_path)
+            key = parsed_url.path.lstrip('/')
+            root_folder = '/'.join(key.split('/')[:3])
+            results_key = f'{root_folder}/results.csv'
+            _upload_dataframe_to_s3(scores, s3_client, bucket, results_key, append=True)
+        else:
+            synth_path = Path(synthesizer_path['synthesizer'])
+            root_path = synth_path.parents[2]
+            result_file = root_path / 'results.csv'
+            _safe_append(scores, result_file)
```
amontanez24, line 741: since there are some places that have these if else splits for local vs remote, should we make classes that represent the ResultsWriter, and then pass either a LocalResultsWriter or AWSResultsWriter to these functions? Then we don't need the if/else blocks. We can just do writer.update_results or whatever functions we pick

sdgym/benchmark.py
```diff
@@ -1080,3 +1218,274 @@ def benchmark_single_table(
...
+    session = boto3.session.Session(
+        aws_access_key_id=s3_client._request_signer._credentials.access_key,
+        aws_secret_access_key=s3_client._request_signer._credentials.secret_key,
+        region_name=s3_client.meta.region_name,
+    )
+    ec2_client = session.client('ec2')
+    print(f'This instance is being created in region: {session.region_name}')  # noqa
+
+    # The user data script
+    user_data_script = textwrap.dedent(f"""\
```
amontanez24, line 1333: I think we should split this part out into its own function as we may override it

sdgym/benchmark.py
```diff
@@ -1080,3 +1218,274 @@ def benchmark_single_table(
...
+
+def _run_on_aws(output_destination, synthesizers, s3_client, job_args_list):
+    bucket_name, job_args_key = _store_job_args_in_s3(output_destination, job_args_list, s3_client)
+    credentials = s3_client._request_signer._credentials
+    access_key = credentials.access_key
+    secret_key = credentials.secret_key
+    region_name = s3_client.meta.region_name
+
+    # Create script content using just the bucket name
+    script_content = f"""
```
amontanez24, line 1292: same thing here about overriding and separating out

sdgym/benchmark.py
```diff
@@ -1080,3 +1218,274 @@ def benchmark_single_table(
...
+                - ``CTGANSynthesizer``
+                - ``CopulaGANSynthesizer``
+                - ``TVAESynthesizer``
+                - ``RealTabFormerSynthesizer``
+        sdv_datasets (list[str] or ``None``):
+            Names of the SDV demo datasets to use for the benchmark. Defaults to
+            ``[adult, alarm, census, child, expedia_hotel_logs, insurance, intrusion, news,
+            covtype]``. Use ``None`` to disable using any sdv datasets.
+        additional_datasets_folder (str or ``None``):
+            The path to an S3 bucket). Datasets found in this folder are
```
amontanez24, line 1432:
```suggestion
            The path to an S3 bucket. Datasets found in this folder are
```

amontanez24, line 1432: I think for now we can make this only be remote

amontanez24 (commented):
It's ok that output_folder is in a different place in terms of order for this function than `benchmark_single_table`

sdgym/utils.py
```diff
@@ -179,3 +185,107 @@ def calculate_score_time(start):
 
     """
     return (get_utc_now() - start).total_seconds()
+
+
+class ResultsWriter(ABC):
```
amontanez24, line 190: I think you can move these to their own file like results_writer.py

sdgym/s3.py
```diff
@@ -144,3 +148,50 @@ def write_csv(data, path, aws_key, aws_secret):
...
+    """Upload a dataframe to S3, optionally appending if the file already exists."""
+    if append:
+        try:
+            response = s3_client.get_object(Bucket=bucket_name, Key=key)
+            existing_csv = response['Body'].read().decode('utf-8')
+            existing_df = pd.read_csv(io.StringIO(existing_csv))
+            data = pd.concat([existing_df, data], ignore_index=True)
+        except botocore.exceptions.ClientError as e:
+            if e.response['Error']['Code'] != 'NoSuchKey':
+                raise e
```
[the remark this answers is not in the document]
amontanez24: @R-Palazzo can we fix this

fealho (changes requested):

sdgym/benchmark.py
```diff
@@ -107,7 +117,62 @@ def _create_detailed_results_directory(detailed_results_folder):
...
+            if match:
+                increments.append(int(match.group(1)))
+
+    return max(increments) + 1 if increments else 1
+
+
+def _setup_output_destination_aws(output_destination, synthesizers, datasets, s3_client):
+    paths = defaultdict(dict)
+    s3_path = output_destination[len(S3_PREFIX) :]
+    parts = s3_path.split('/')
```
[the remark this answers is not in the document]
R-Palazzo: Good catch, addressed in [36b8dea](https://github.com/sdv-dev/SDGym/pull/417/commits/36b8deae1097688c1179703e51615ed7be3e6fdf)

amontanez24 (approved):

sdgym/s3.py
```diff
@@ -144,3 +150,52 @@ def write_csv(data, path, aws_key, aws_secret):
     """
     data_contents = data.to_csv(index=False).encode('utf-8')
     write_file(data_contents, path, aws_key, aws_secret)
+
+
+def _parse_s3_uri(s3_uri):
```
fealho, line 155: Isn't this method the same as `parse_s3_path` which already exists?

fealho (commented):
I left one final comment, it should be good to go after that

fealho (approved):

PR #420, opened 2025-07-07
Add SDGymResultsExplorer class

### LINKED ISSUE #411: Add SDGymResultsExplorer class
### Problem Description

_As a user, I'd like to easily navigate the results of an SDGym run. This means being able to access the synthesizers, synthetic data and scores easily._

### Expected behavior

Add the following class:

```python
class SDGymResultsExplorer:
    def __init__(self, path):
        """
        Args:
            path (str): Either the root folder where the results were stored or the s3 bucket name.
        """
        pass

    def list(self):
        """Lists all of the results in the bucket.
        Returns:
            list: List of all run names. Eg. ['SDGym_results_04_18_2025', 'SDGym_results_05_24_2025', 'SDGym_results_06_28_2025']
        """
        pass

    def load_synthesizer(self, results_folder_name=`SDGym_run_04_18_2025`, dataset_name='adult', synthesizer_name='ctgan'):
        """Load the synthesizer object.
        Args:
            run_folder_name (str): The name of the folder for the desired run.
            dataset_name (str): The name of the dataset.
            synthesizer_name (str): The name of the synthesizer (eg. ctgan).
        Returns:
            sdgym.synthesizers.BaselineSynthesizer
        """
        pass

    def load_synthetic_data(self, results_folder_name=`SDGym_run_04_18_2025`, dataset_name='adult', synthesizer_name='ctgan'):
        """Load the synthetic data created by a specific synthesizer for a given run and dataset.
        Args:
            run_folder_name (str): The name of the folder for the desired run.
            dataset_name (str): The name of the dataset.
            synthesizer_name (str): The name of the synthesizer (eg. ctgan).
        Returns:
            pd.DataFrame
        """
        pass
    
    def load_real_data(self, dataset_name='adult'):
        """Load the data used for fitting.
        Args:
            dataset_name (str): The name of the dataset.
        Returns:
            pd.DataFrame
        """
        pass
```
- This class can be in a module called `sdgym_results_explorer`

### Additional context

- This issue is dependent on #410

R-Palazzo:
Resolve #411
CU-86b5dy740

I added `aws_access_key_id` and `aws_secret_access_key` parameters to the `SDGymResultsExplorer` so it can access s3 buckets.
One can try on aws with this output_destination as example:
`output_destination = 's3://sdgym-benchmark/Debug/Issue_414_test_8'`

@amontanez24, what are the expectations for `load_real_data()` for a dataset that is not from SDV?

### DIFF
--- a/sdgym/sdgym_result_explorer/result_handler.py
+++ b/sdgym/sdgym_result_explorer/result_handler.py
@@ -18,7 +18,7 @@ class ResultsHandler(ABC):
         pass
 
     @abstractmethod
-    def get_file_path(self, results_folder_name, dataset_name, synthesizer_name, type):
+    def get_file_path(self, path_parts, end_filename):
         """Validate access to a specific file in the results directory."""
         pass
 


sdgym/sdgym_result_explorer/result_explorer.py
```diff
@@ -0,0 +1,194 @@
...
+        if dataset_name in DEFAULT_DATASETS:
+            dataset_path = get_dataset_paths(
+                datasets=[dataset_name],
+                aws_key=self.aws_access_key_id,
+                aws_secret=self.aws_secret_access_key,
+            )[0]
+
+        data, _ = load_dataset(
+            'single_table',
+            dataset_path,
```
fealho, line 189: dataset_path is not set when dataset_name is not in DEFAULT_DATASETS

sdgym/sdgym_result_explorer/result_explorer.py
```diff
@@ -0,0 +1,194 @@
...
+    def load_synthesizer(
+        self,
+        results_folder_name='SDGym_run_04_18_2025',
+        dataset_name='adult',
+        synthesizer_name='ctgan',
+    ):
+        """Load the synthesizer object.
+
+        Args:
+            run_folder_name (str): The name of the folder for the desired run.
```
fealho, line 120: results_folder_name

sdgym/sdgym_result_explorer/result_explorer.py
```diff
@@ -0,0 +1,194 @@
...
+    def load_synthetic_data(
+        self,
+        results_folder_name='SDGym_run_04_18_2025',
+        dataset_name='adult',
+        synthesizer_name='ctgan',
+    ):
+        """Load the synthetic data created by a specific synthesizer for a given run and dataset.
+
+        Args:
+            run_folder_name (str): The name of the folder for the desired run.
```
fealho, line 150: results_folder_name

sdgym/sdgym_result_explorer/result_explorer.py
```diff
@@ -0,0 +1,194 @@
...
+        file_path = self._validate_access(
+            results_folder_name, dataset_name, synthesizer_name, 'synthesizer'
+        )
+        if self._is_s3_path:
+            response = self._s3_client.get_object(
+                Bucket=self._bucket_name, Key=f'{self._prefix}{file_path}'
+            )
+            synthesizer = pickle.loads(response['Body'].read())
+        else:
+            with open(os.path.join(self.path, file_path), 'rb') as f:
```
fealho, line 136: Would this work on Windows? Because `file_path = f'{results_folder_name}/{dataset_name}_{date}/{synthesizer_name}/{end_filename}'`, while `self.path=C:\Users\data\results` or something like that.

sdgym/sdgym_result_explorer/result_handler.py
```diff
@@ -0,0 +1,110 @@
...
+    def __init__(self, path, aws_key, aws_secret):
+        self.s3_client = boto3.client(
+            's3',
+            aws_access_key_id=aws_key,
+            aws_secret_access_key=aws_secret,
+        )
+        self.bucket_name = path.split('/')[2]
+        self.prefix = '/'.join(path.split('/')[3:]).rstrip('/') + '/'
+
+    def list_runs(self):
```
amontanez24, line 82: I'm wondering if we should call this list runs or list something else. In #415 , we are calling a run a unique call of the benchmark function. Here, we are referring to all the runs for a date I believe. Maybe we should have a different name for that concept.

sdgym/sdgym_result_explorer/result_explorer.py
```diff
@@ -0,0 +1,88 @@
...
+from sdgym.benchmark import DEFAULT_DATASETS
+from sdgym.datasets import get_dataset_paths, load_dataset
+from sdgym.s3 import _get_s3_client, is_s3_path
+from sdgym.sdgym_result_explorer.result_handler import LocalResultsHandler, S3ResultsHandler
+
+
+def _validate_path(path, aws_access_key_id=None, aws_secret_access_key=None):
+    """Validates the provided path to ensure it is either a local directory or an S3 bucket."""
+    if is_s3_path(path):
+        s3_client = _get_s3_client(path, aws_access_key_id, aws_secret_access_key)
```
amontanez24, line 14: I think this was confusing because it is validating but also being used to determine if the path ios local or on s3. Can we just determine if it's an s3 path, and if so then get the client all in the init. This function can just be a validation function for local paths

fealho (commented):
Looks pretty good.

fealho (approved):

amontanez24 (approved):

PR #433, opened 2025-08-11
remove _synthesizer

### LINKED ISSUE #430: Make the synthesizer names consistent throughout SDGym
### Problem Description
In the upcoming SDGym release, the benchmarking function will be saving the intermediary artifacts (synthesizers and synthetic data) into the detailed results folder. To allow easy, programmatic navigation of this folder, we are also introducing a new class to explore results.

See: https://github.com/sdv-dev/SDGym/issues/411 and https://github.com/sdv-dev/SDGym/issues/410

The problem is that the new functionality doesn't use consistent terminology to refer to the synthesizers.

- **Existing terminology**: The synthesizers are referred to by their full class name, as defined in SDV or other libraries. For example, [this synthesizer](https://docs.sdv.dev/sdv/single-table-data/modeling/synthesizers/ctgansynthesizer) is referred to as `'CTGANSynthesizer'` when (a) setting up the benchmark/choosing synthesizers, as well as (b) in the final [results summary](https://docs.sdv.dev/sdgym/benchmarking/interpreting-results)
- **New functionality**: The new functionality refers to the synthesizers by a shorthand name. For example, the CTGANSynthesizer is shortened to the string `'ctgan'`. This happens when (a) saving the synthesizer, (b) saving the synthetic data, and (c) when using the results explorer object to retrieve those objects. 

### Expected behavior
The synthesizers should always be referred to by the same name within SDGym.  We should align with the **existing terminology** that has already been established in SDGym.

Changes to be made:

1. Save the synthesizers using the full class name. Because this will already have `Synthesizer` within the name, there is no need to add `_synthesizer` to it.
    - For example, save the object as `CTGANSynthesizer.pkl` (rather than `ctgan_synthesizer.pkl` as it's implemented now)
2. Save the synthetic data using the full class name too. Continue to append `_synthetic_data` to that name
    - For example, save the synthetic data as `CTGANSynthesizer_synthetic_data.csv` (rather than `ctgan_synthetic_data.csv` as it's implemented now)
3. The results explorer should be able to recall those artifacts using the same full name as used above.

```python
synthesizer = results_explorer.load_synthesizer(
    results_folder_name=`SDGym_run_04_18_2025`,
    dataset_name='adult',
    synthesizer_name='CTGANSynthesizer' # not 'ctgan'
)

synthetic_data = results_explorer.load_synthetic_data(
    results_folder_name=`SDGym_run_04_18_2025`,
    dataset_name='adult',
    synthesizer_name='CTGANSynthesizer' # not 'ctgan'
)
```

R-Palazzo:
Resolve #430
CU-86b624bjr

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -163,7 +163,7 @@ def _setup_output_destination_aws(output_destination, synthesizers, datasets, s3
             synth_folder = f'{dataset_folder}/{synth_name}'
             s3_client.put_object(Bucket=bucket_name, Key=synth_folder + '/')
             paths[dataset][synth_name] = {
-                'synthesizer': f's3://{bucket_name}/{synth_folder}/{synth_name}_synthesizer.pkl',
+                'synthesizer': f's3://{bucket_name}/{synth_folder}/{synth_name}.pkl',
                 'synthetic_data': f's3://{bucket_name}/{synth_folder}/{synth_name}_synthetic_data.csv',
                 'benchmark_result': f's3://{bucket_name}/{synth_folder}/{synth_name}_benchmark_result.csv',
                 'results': f's3://{bucket_name}/{top_folder}/results_{today}_{increment}.csv',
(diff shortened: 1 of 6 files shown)


sdv-team:
Task linked: [CU-86b624bjr SDGym - Make the synthesizer names consistent throughout SDGym #430](https://app.clickup.com/t/86b624bjr)

rwedge (approved):

Issue #438, opened 2025-09-02
For graceful degradation purposes, always run the UniformSynthesizer on the requested datasets


npatki:
### Problem Description
Depending on the synthesizers that are used in SDGym, it is possible that the final results contain missing values -- for example, if a synthesizer crashes or times out during fitting/sampling. 

However, the overall results are meant to showcase what an actual user may do when using such a synthesizer. In this case, leaving the values completely blank is not necessarily what they'd do. A graceful degradation would be to create synthetic data with a _basic_ synthesizer that can get the job done, but with very minimal intelligence. We should use the [UniformSynthesizer](https://docs.sdv.dev/sdgym/customization/synthesizers/basic-synthesizers) for graceful degradation. 

### Expected behavior
1. Make sure that the `UniformSynthesizer` is always run -- regardless of whether or not it is explicitly passed into the parameters of SDGym.
2. In the list of [default synthesizers](https://docs.sdv.dev/sdgym/benchmarking/running-a-benchmark#setup) that are run by the benchmarking framework, add the UniformSynthesizer.

Note that UniformSynthesizer will behave just like any other synthesizer that is requested to run -- i.e. an artifact should be saved for it, its results should appear in the final summary etc.

### Additional context
To update the missing values, we'll file a new issue.

PR #443, opened 2025-09-24
Use fallback synthesizer to get adjusted scores

### LINKED ISSUE #439: To simulate graceful degradation, fallback to using the results from the UniformSynthesizer
### Problem Description
Depending on the synthesizers that are used in SDGym, it is possible that the final results contain missing values -- for example, if a synthesizer crashes or times out during fitting/sampling.

However, the overall results are meant to showcase what an actual user may do when using such a synthesizer. In this case, leaving the values completely blank is not necessarily what they'd do. A graceful degradation would be to create synthetic data with a basic synthesizer that can get the job done, but with very minimal intelligence. In #438, we are running the `UniformSynthesizer` for graceful degradation. 

### Expected behavior
Issue #438 ensures that the UniformSynthesizer will always be there during a benchmarking run. This issue is to incorporate the results from UniformSynthesizer into the final, [summarized results file](https://docs.sdv.dev/sdgym/benchmarking/interpreting-results).

To the final, summarized results CSV, add the following columns:

1. `adjusted_total_time`, whose calculation is as follows:
    - If everything ran fine: adjusted total time = uniform fit time + synthesizer fit time + synthesizer sample time. _This is because for graceful degradation, a UniformSynthesizer must always be fit, even if it's not used._
   - If there was an crash that happened before the timeout was reached: adjusted total time = uniform fit time + synthesizer fit time (if available) + sample time (if available) + uniform sample time. _The UniformSynthesizer is always fit, and then sampled from the point of time of the crash._
   - If there was a timeout: adjusted total time = timeout value + uniform fit time + uniform sample time. _The synthesizer is allowed to run to completion before the UniformSynthesizer is used._
2. `adjusted_quality_score`, whose calculation is as follows:
    - If everything ran fine: adjusted_quality_score = quality_score
   - If there is any kind of error (crash or timeout): adjusted_quality_score = quality score of the UniformSynthesizer. _We gracefully degrade to the UniformSynthesizer's synthetic data._

rwedge:
Resolves #439 

CU-86b6h454g

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -787,6 +787,7 @@ def _run_jobs(multi_processing_config, job_args_list, show_progress, result_writ
         raise SDGymError('No valid Dataset/Synthesizer combination given.')
 
     scores = pd.concat(scores, ignore_index=True)
+    _add_adjusted_scores(scores=scores, timeout=job_args_list[0][5])
     output_directions = job_args_list[0][-2]
     result_writer = job_args_list[0][-1]
     if output_directions and result_writer:
@@ -810,12 +811,14 @@ def _get_empty_dataframe(
         'Synthesizer_Size_MB': [],
         'Sample_Time': [],
         'Evaluate_Time': [],
+        'Adjusted_Total_Time': [],
     })
 
     if compute_diagnostic_score:
         scores['Diagnostic_Score'] = []
     if compute_quality_score:
         scores['Quality_Score'] = []
+        scores['Adjusted_Quality_Score'] = []
     if compute_privacy_score:
         scores['Privacy_Score'] = []
     if sdmetrics:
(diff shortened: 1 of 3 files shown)


sdgym/benchmark.py
```diff
@@ -1051,6 +1054,46 @@ def _ensure_uniform_included(synthesizers):
...
+def _add_adjusted_scores(scores, timeout):
+    uniform_row = scores[scores['Synthesizer'] == 'UniformSynthesizer'].iloc[0]
+    uniform_fit_time = uniform_row['Train_Time']
+    uniform_sample_time = uniform_row['Sample_Time']
+    if 'Quality_Score' in scores.columns:
+        uniform_quality_score = uniform_row['Quality_Score']
+
+    def _get_adjusted_time(row):
+        if pd.isna(uniform_fit_time) or pd.isna(uniform_sample_time):
+            return
```
amontanez24, line 1066: does this fill with nan if you just return?

sdgym/benchmark.py
```diff
@@ -1051,6 +1054,46 @@ def _ensure_uniform_included(synthesizers):
...
+                adjusted_total_time += timeout
+            else:
+                if not pd.isna(row['Train_Time']):
+                    adjusted_total_time += row['Train_Time']
+                if not pd.isna(row['Sample_Time']):
+                    adjusted_total_time += row['Sample_Time']
+            adjusted_total_time += uniform_sample_time
+        else:
+            adjusted_total_time += row['Train_Time']
+            adjusted_total_time += row['Sample_Time']
```
amontanez24, line 1079: if there is no error, do we still need to add the uniform_fit_time? Isn't this a case where no adjustment is needed?

rwedge: for graceful degradation, a UniformSynthesizer must always be fit, even if it's not used, so the fit time is included even if it wasn't needed

amontanez24: ah I see

sdgym/benchmark.py
amontanez24: that's fine

sdgym/benchmark.py
```diff
@@ -1051,6 +1054,49 @@ def _ensure_uniform_included(synthesizers):
         synthesizers.append(UniformSynthesizer)
 
 
+def _add_adjusted_scores(scores, timeout):
+    try:
+        uniform_row = scores[scores['Synthesizer'] == 'UniformSynthesizer'].iloc[0]
+    except IndexError:
+        uniform_row = pd.Series({'Train_Time': None, 'Sample_Time': None, 'Quality_Score': None})
+    uniform_fit_time = uniform_row['Train_Time']
```
R-Palazzo, line 1062: Add a blank line

tests/integration/test_benchmark.py
```diff
@@ -672,7 +698,8 @@ def test_benchmark_single_table_with_output_destination(tmp_path):
         f'{output_destination}/SDGym_results_{today_date}/results_{today_date}_1.csv'
     )
     pd.testing.assert_frame_equal(results, saved_result, check_dtype=False)
-    pd.testing.assert_frame_equal(results, score_saved_separately, check_dtype=False)
+    results_no_adjusted = results.drop(columns=['Adjusted_Total_Time', 'Adjusted_Quality_Score'])
```
R-Palazzo, line 701: Should the `score_saved_separately` also include the `Adjusted_Total_Time` and `Adjusted_Quality_Score`? I think it works if the very first step of the benchmark is to run the UniformSynthesizer, then when any synthesizer is saved, even independently, it could have those two columns.

R-Palazzo (commented):
Looking really good!

Is this in this Issue that UniformSyntheiszer must run first before any other synthesizer? Or should the UniformSynthesizer be included in the list of default synthesizers?
@amontanez24

rwedge:
@amontanez24 @R-Palazzo Always running UniformSynthesizer has already been handled by #438. The outstanding question I see is if the saved scores of individual jobs should include the adjusted columns or if only the combined results csv needs to include the adjusted scores.

amontanez24 (approved):

PR #472, opened 2025-10-31
Add `DatasetExplorer` class that provides a summary of all datasets in a bucket

### LINKED ISSUE #469: Add a `DatasetExplorer` class that provides a summary of all datasets in a bucket (for a given modality)
**Due By**: Fri Oct 31 (9AM Boston time), so that we can release later that same day Oct 31

### Problem Description
Currently, there are many different datasets that are available for SDGym to benchmark on. Users can also input a private S3 bucket if they want to test on additional datasets.

However, besides benchmarking on those datasets, there is no way for users to (a) see what those dataset are, and (b) get an understanding of the _attributes_ of those datasets. Attributes are extra information about the data schema, for example, the total # of tables, the size of the data, the depth of the schema, etc.

### Expected behavior
Add `DatasetExplorer` class to SDGym. 

**__init__**: The 3 init parameters are all optional:
- `s3_url`: The URL of the S3 bucket that contains the dataset. By default, this will be the _new_ public S3 bucket that is offered by DataCebo.
- `aws_access_key_id`: A string containing the AWS access key id. This is needed if the `s3_url` is for a private S3 bucket.
- `aws_secret_access_key`: A string containing the AWS secret access key. This is needed if the `s3_url` is for a private S3 bucket.

```python

de = DatasetExplorer(
  s3_url='s3://my_bucket/',
  aws_secret_access_key='my_secret'
)
```

Then add a function called **summarize_datasets** that summarizes the datasets inside the bucket. It takes the following parameters:
- (required) `modality`: The modality of the data to summarize. This should be one of `'single_table'`, `'multi_table'` or `'sequential'`.
- `output_filepath`: A string with the full output filepath where the results will be written. This should end in `.csv`

The function should output a pandas.DataFrame object that contains the results summary. If the `output_filepath` is provided, then the same dataframe would be written as a CSV in the file provided. (See below for the format of the object.)

```python
dataset_summary = de.summarize_datasets(
  modality='single_table',
  output_filepath='datasets_summary.csv')
```

### Dataset Summary
The dataset summary should contain 1 row per dataset that is in the bucket for the given modality. It should contain the following columns:

- `Dataset`: The name of the dataset
- `Datasize_Size_MB`: The overall size of the dataset in MB (this can be read from the metainfo.yaml file)
- `Num_Tables`: The total number of tables in the dataset
- `Total_Num_Columns`: The total number of columns (sum of all the column counts across all tables)
- `Total_Num_Columns_Categorical`: The total number of columns that are listed sdtype `categorical` in the metadata (sum of the column counts across all tables)
- `Total_Num_Columns_Numerical`: The total number of columns that are listed sdtype `numerical` in the metadata (sum of the column counts across all tables)
- `Total_Num_Columns_Datetime`: The total number of columns that are listed sdtype `datetime` in the metadata (sum of the column counts across all tables)
- `Total_Num_Columns_PII`: The total number of columns that are pii in the metadata (sum of the column counts across all tables). _Note that by default, PII is assumed to be true for all sdtypes that are not numerical, categorical, boolean, datetime, or id._
- `Total_Num_Columns_ID_NonKey`: The total number of columns that are listed as sdtype `id` in the metadata _and_ are not any type of key (primary, foreign, alternate, sequence). (sum of the column counts across all tables)
- `Max_Num_Columns_Per_Table`: The maximum number of columns that a table has in the dataset
- `Total_Num_Rows`: The total number of rows in the schema (sum of row counts across all tables)
- `Max_Num_Rows_Per_Table`: The maximum number of rows in any one table 
- `Num_Relationships`: The total # of relationships that are listed in the metadata
- `Max_Schema_Depth`: The maximum depth of the schema. _Note that max schema depth is the max # of tables in the chain. So a single parent-child connection is considered depth=2 NOT 1._
- `Max_Schema_Branch`: The maximum number of foreign keys that point to a given table. _So if a parent table has 3 child tables that are pointing to it, its branching factor is 3. We want the maximum of any table._

### Additional Context
This issue is to be done after #468, in which we'll update the structure of the S3 bucket.

pvk-developer:
Resolves #469 
CU-86b78wcqv

### DIFF
--- a/sdgym/dataset_explorer.py
+++ b/sdgym/dataset_explorer.py
@@ -2,6 +2,7 @@
 
 from collections import defaultdict
 from pathlib import Path
+from urllib.parse import urlparse
 
 import pandas as pd
 from sdv.metadata import Metadata
@@ -29,6 +30,7 @@ class DatasetExplorer:
         self.s3_url = s3_url
         self.aws_access_key_id = aws_access_key_id
         self.aws_secret_access_key = aws_secret_access_key
+        self._bucket_name = urlparse(self.s3_url).netloc
 
     @staticmethod
     def _get_max_schema_branch_factor(relationships):
@@ -192,9 +194,10 @@ class DatasetExplorer:
                 for an individual dataset.
         """
         results = []
+
         datasets = _get_available_datasets(
             modality=modality,
-            bucket=self.s3_url,
+            bucket=self._bucket_name,
             aws_access_key_id=self.aws_access_key_id,
             aws_secret_access_key=self.aws_secret_access_key,
         )
@@ -205,7 +208,7 @@ class DatasetExplorer:
             data, metadata_dict = load_dataset(
                 modality,
                 dataset=dataset_name,
-                bucket=self.s3_url,
+                bucket=self._bucket_name,
                 aws_access_key_id=self.aws_access_key_id,
                 aws_secret_access_key=self.aws_secret_access_key,
             )
--- a/tests/unit/test_dataset_explorer.py
+++ b/tests/unit/test_dataset_explorer.py
@@ -224,14 +224,14 @@ class TestDatasetExplorer:
         # Assert
         mock_get_datasets.assert_called_once_with(
             modality='single_table',
-            bucket=BUCKET,
+            bucket='sdv-datasets-public',
             aws_access_key_id=None,
             aws_secret_access_key=None,
         )
         mock_load_dataset.assert_called_once_with(
             'single_table',
             dataset='test',
-            bucket=BUCKET,
+            bucket='sdv-datasets-public',
             aws_access_key_id=None,
             aws_secret_access_key=None,
         )


sdgym/datasets.py
```diff
@@ -312,7 +312,12 @@ def get_dataset_paths(
                 if _path_contains_data_and_metadata(dataset) and dataset not in datasets:
                     datasets.append(dataset)
         else:
-            datasets = _get_available_datasets(modality, bucket=bucket)
+            datasets = _get_available_datasets(
+                modality,
+                bucket=bucket,
+                aws_access_key_id=aws_access_key_id,
+                aws_secret_access_key=aws_secret_access_key
```
pvk-developer: Discovered this bug while bug hunting on collab. Locally was hard to spot since the default `s3` client would use my private credentials.

tests/integration/test_dataset_explorer.py
```diff
@@ -0,0 +1,44 @@
...
+    'Total_Num_Rows',
+    'Max_Num_Rows_Per_Table',
+    'Num_Relationships',
+    'Max_Schema_Depth',
+    'Max_Schema_Branch',
+]
+
+
+@pytest.mark.parametrize('modality', ['single_table', 'multi_table', 'sequential'])
+def test_end_to_end_dataset_explorer(modality, tmp_path):
```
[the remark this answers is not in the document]
pvk-developer: It takes a bit depending on your connection. The main time consumption is on downloading the datasets.

sdgym/dataset_explorer.py
```diff
@@ -0,0 +1,270 @@
...
+            )
+            for column_name, column in table.columns.items():
+                sdtype = column['sdtype']
+                if sdtype in ['categorical', 'boolean']:
+                    results['Total_Num_Columns_Categorical'] += 1
+                elif sdtype in ['numerical']:
+                    results['Total_Num_Columns_Numerical'] += 1
+                elif sdtype in ['datetime']:
+                    results['Total_Num_Columns_Datetime'] += 1
+                elif sdtype in ['id'] and column_name != table.primary_key:
```
fealho: The documentation mentions it shouldn't be any type of key, not just primary key.

fealho (approved):
Looks good! Only two small comments.

sdgym/dataset_explorer.py
pvk-developer: Done here [40b0b83](https://github.com/sdv-dev/SDGym/pull/472/commits/40b0b83e54042bf86591f507711826b393c392ea)

sdgym/dataset_explorer.py
```diff
@@ -0,0 +1,274 @@
...
+        def dfs(table):
+            if table not in child_map or not child_map[table]:
+                return 1
+
+            return 1 + max(dfs(child) for child in child_map[table])
+
+        parent_map = metadata._get_parent_map()
+        root_tables = [table for table in child_map.keys() if table not in parent_map]
+        if not root_tables:
+            return 1
```
amontanez24: what is the situation where there are no root tables but there is a child_map? In other words shouldn't line 67 cover this?

pvk-developer: Yeah actually this is because I messed up with the unit tests and added that line after it, the test looks like this:
```python
    @patch('sdgym.dataset_explorer.Metadata.load_from_dict')
    def test_get_metadata_summary_with_dict(self, mock_load_from_dict):
        """Ensure that the ``get_metadata_summary`` method converts metadata dict to Metadata."""
        # Setup
        mock_metadata = MagicMock()
        mock_metadata.relationships = []
        mock_metadata.tables = {}
        mock_load_from_dict.return_value = mock_metadata

        # Run
        result = DatasetExplorer.get_metadata_summary({})

        # Assert
        mock_load_from_dict.assert_called_once()
        assert isinstance(result, dict)
```
So everything is a mock until `root_tables` that becomes empty (so this is covering a pretty edge case, which I agree on a valid metadata it will never come). Should I remove it ?

PR #489, opened 2025-11-07
SDGym should be able to automatically discover SDV Enterprise synthesizers

### LINKED ISSUE #481: SDGym should be able to automatically discover SDV Enterprise synthesizers
### Problem Description
SDGym is designed to be able to benchmark synthesizers. SDV synthesizers are natively supported, while external, 3rd party synthesizers can be integrated [as a custom synthesizer](https://docs.sdv.dev/sdgym/customization/synthesizers/custom-synthesizers).

Currently, the main publicly-available SDV synthesizers are natively supported in SDGym. However, this does not apply to any synthesizers in SDV Enterprise or bundles (eg. the [SegmentSynthesizer](https://docs.sdv.dev/sdv/single-table-data/modeling/synthesizers/segmentsynthesizer) from the [XSynthesizers bundle](https://docs.sdv.dev/sdv/explore/sdv-bundles/xsynthesizers)). The user would have to integrate these manually as custom synthesizers.

### Expected behavior
I expect that SDGym should be able to automatically discover **any single- or multi-table SDV synthesizer** that I have access to in my Python environment. 

**Single table synthesizers**: SDGym should be able to search for the synthesizer name in the `sdv.single_table` namespace. If that synthesizer name exists, then it should be able to load it into the appropriate format (eg. see [this base class](https://github.com/sdv-dev/SDGym/blob/main/sdgym/synthesizers/sdv.py#L14)). As a result, a user should be able to benchmark any SDV single-table synthesizer that they have access to by providing the string of its name.

```python
# assuming I have SDV Enterprise and the XSynthesizers bundle already installed in my environment
# I should be able to benchmark the synthesizers by inputting their names
sdgym.benchmark_single_table(
    synthesizers=['SegmentSynthesizer', 'XGCSynthesizer'] 
)
```

Additionally, I should be able to create variants of these types of synthesizers [following this guide](https://docs.sdv.dev/sdgym/customization/synthesizers/sdv-synthesizers#create_sdv_synthesizer_variant).
```python
from sdgym import create_sdv_synthesizer_variant

BiSegmentSyntheiszer = create_sdv_synthesizer_variant(
  synthesizer_class='SegmentSynthesizer',
  synthesizer_parameters={ 'n_segments': 2 }
  display_name='BiSegmentSynthesizer'
)
```

**Multi table synthesizers**: SDGym should also be able to search for synthesizer names in the `sdv.multi_table` namespace. In this case, it should be able to find and load in the synthesizer in a similar format to single-table but with some modifications:
- `_get_trained_synthesizer` function: The input parameter `data` should be a dictionary of dataframes instead of a single pandas DataFrame
- `_sample_from_synthesizer` function: Instead of `n_samples`, the parameter it accepts should be `scale`.

_Note that multi-table benchmarking is not currently supported in SDGym but we hope to add support for it in the future._

### Additional context
We should also generally [clean up this file](https://github.com/sdv-dev/SDGym/blob/main/sdgym/synthesizers/sdv.py#L1) when making the changes.
- The nomenclature is "single-table" and "multi-table". Get rid of any references to "tabular" or "relational" (these were the old names)
- None of the SDV synthesizers should be hard-coded (eg. GaussianCopulaSynthesizer, CTANSynthesizers, etc.). All of them should be able to be dynamically discovered from `sdv.single_table` and `sdv.multi_table` modules
- We can remove the sequential code for now.

### LINKED ISSUE #491: Rename create_sdv_synthesizer_variant to create_synthesizer_variant
### Problem Description

The current function name `create_sdv_synthesizer_variant` suggests it only supports SDV synthesizers. However, it currently supports any other SDGym-native synthesizers, including basic and third-party ones such as `RealTabFormerSynthesizer`. So we should rename this method to `create_synthesizer_variant` to clarify that any SDGym-supported synthesizer can have variants using this method.

### Expected behavior

Rename the function to `create_synthesizer_variant`

R-Palazzo:
Resolve #481
Resolve #491 
CU-86b7b0kaa
I share this [notebook](https://colab.research.google.com/drive/1fOS8oRtxfigsK8Yd0qeLS8m4RQ0XWxsx?usp=sharing) to try out the feature.

### DIFF
--- a/sdgym/synthesizers/generate.py
+++ b/sdgym/synthesizers/generate.py
@@ -14,9 +14,6 @@ def create_synthesizer_variant(display_name, synthesizer_class, synthesizer_para
             Name of the SDV synthesizer class to wrap.
         synthesizer_parameters (dict):
             A dictionary of the parameter names and values that will be used for the synthesizer.
-        modality (str):
-            The modality of the synthesizer, either 'single_table' or 'multi_table'.
-            Defaults to 'single_table'.
 
     Returns:
         class:
--- a/sdgym/synthesizers/base.py
+++ b/sdgym/synthesizers/base.py
@@ -36,7 +36,7 @@ class BaselineSynthesizer(abc.ABC):
     @classmethod
     def _get_supported_synthesizers(cls):
         """Get the natively supported synthesizer class names."""
-        subclasses = cls.get_subclasses(include_parents=False)
+        subclasses = cls.get_subclasses(include_parents=True)
         synthesizers = set()
         for name, subclass in subclasses.items():
             if subclass._NATIVELY_SUPPORTED:


tests/integration/test_benchmark.py
```diff
@@ -54,7 +54,7 @@ def test_benchmark_single_table_basic_synthsizers():
     ] == quality_scores.index.tolist()
 
 
-@pytest.mark.skipif(sys.platform.startswith('darwin'), reason='Test not supported on github MacOS')
+@pytest.mark.skip('`create_sdv_synthesizer_variant` should only work with SDV synthesizers.')
```
R-Palazzo: Let me know if this makes sense. In this test, we were creating a variant of RealTabFormer with `create_sdv_synthesizer_variant`, which I found confusing.

sdgym/synthesizers/generate.py
```diff
@@ -92,61 +71,28 @@ def create_single_table_synthesizer(
...
     class NewSynthesizer(BaselineSynthesizer):
-        """New Synthesizer class.
-
-        Args:
-            get_trained_synthesizer_fn (callable):
-                Function to replace the ``get_trained_synthesizer`` method.
-            sample_from_synthesizer_fn (callable):
-                Function to replace the ``sample_from_synthesizer`` method.
-        """
-
```
pvk-developer: Why are all this docstrings removed ?

sdgym/synthesizers/generate.py
```diff
@@ -219,69 +167,4 @@ def sample_from_synthesizer(self, synthesizer):
...
-            the results are generated
-        get_trained_synthesizer_fn (callable):
-            A function to generate and train a synthesizer, given the real data and metadata.
-        sample_from_synthesizer (callable):
-            A function to sample from the given synthesizer.
-
-    Returns:
-        class:
-            The synthesizer class.
-    """
```
pvk-developer: Why are we removing all this ?

R-Palazzo: The last point of the issue mentions removing the sequential code for now

pvk-developer:
> @amontanez24 @pvk-developer Should we support using `DayZSynthesizer` in a benchmark? Right now it won’t be picked up, because our discovery only looks for subclasses of `BaseSingleTableSynthesizer` and `BaseMultiTableSynthesizer`, and DayZ doesn’t inherit from those. That’s a small change to make.
> 
> Two things to consider:
> 
> * DayZSynthesizer does not have a `fit()` method; its SDGym wrapper should take this into account to make it work in a benchmark.
> * Both the `single-table` and `multi-table` versions are named `DayZSynthesizer`, but based on the context we should still be able to determine which one we need I think

If I had to write this as per the issue is saying, we should get all the synthesizers that are accessed by `sdv.single_table` such as:
```python
In [3]: from sdv import single_table

In [4]: dict([(name, cls) for name, cls in single_table.__dict__.items() if isinstance(cls, type)])
Out[4]: 
{'CopulaGANSynthesizer': sdv.single_table.copulagan.CopulaGANSynthesizer,
 'GaussianCopulaSynthesizer': sdv.single_table.copulas.GaussianCopulaSynthesizer,
 'CTGANSynthesizer': sdv.single_table.ctgan.CTGANSynthesizer,
 'TVAESynthesizer': sdv.single_table.ctgan.TVAESynthesizer,
 'DayZSynthesizer': sdv_enterprise.sdv.single_table.dayz.day_zero.DayZSynthesizer,
 'BootstrapSynthesizer': sdv_enterprise.sdv.single_table.bootstrap.bootstrap.BootstrapSynthesizer,
 'DPGCFlexSynthesizer': sdv_enterprise.sdv.single_table.differential_privacy.dp_gc_flex_synthesizer.DPGCFlexSynthesizer,
 'DPGCSynthesizer': sdv_enterprise.sdv.single_table.differential_privacy.dp_gc_synthesizer.DPGCSynthesizer,
 'SegmentSynthesizer': sdv_enterprise.sdv.single_table.segment.segment.SegmentSynthesizer,
 'XGCSynthesizer': sdv_enterprise.sdv.single_table.xgc.xgc.XGCSynthesizer}
```
Then my other comment on how you got the subclasses can go away and you could just use this functionality instead.

sdgym/benchmark.py
```diff
@@ -907,8 +907,7 @@ def _create_sdgym_script(params, output_filepath):
 from io import StringIO
 import sdgym
 from sdgym.synthesizers.sdv import (CopulaGANSynthesizer, CTGANSynthesizer,
-    GaussianCopulaSynthesizer, HMASynthesizer, PARSynthesizer, SDVRelationalSynthesizer,
-    SDVTabularSynthesizer, TVAESynthesizer)
+    GaussianCopulaSynthesizer, HMASynthesizer, PARSynthesizer, TVAESynthesizer)
```
pvk-developer: this will fail now

sdgym/synthesizers/generate.py
```diff
@@ -1,95 +1,63 @@
-"""Synthesizers module."""
+"""Helpers to create SDV synthesizer variants."""
```
pvk-developer: If this module is `generate.py` then either this description is wrong (because we have RealTabFormer in here too) or the functions you added don't belong here.

I believe this is `Helpers to create SDGym synthesizer variants`, that may be more appropriate.

sdgym/synthesizers/sdv.py
```diff
@@ -1,126 +1,94 @@
...
+        _validate_inputs(sdv_name, modality, parameters)
+        self.sdv_name = sdv_name
+        self.modality = modality
+        self.parameters = parameters or {}
 
-class PARSynthesizer(SDVTimeseriesSynthesizer):
-    """Model wrapping the ``PARSynthesizer`` model."""
+    def __repr__(self):
+        """Define the string representation of the synthesizer."""
+        return f'<{self.__class__.__name__} sdv_name={self.sdv_name!r} modality={self.modality!r}>'
```
pvk-developer: > The goal of `__repr__` is to be unambiguous and, if possible, valid Python code that could recreate the object.

I would suggest changing this to something like:
```python
    return f'{self.__class__.__name__}(sdv_name={self.sdv_name!r}, modality={self.modality!r})'
```

sdgym/utils.py
```diff
@@ -47,19 +48,31 @@ def get_synthesizers(synthesizers):
...
         if isinstance(synthesizer, str):
-            baselines = BaselineSynthesizer.get_subclasses(include_parents=True)
-            if synthesizer in baselines:
-                LOGGER.info('Trying to import synthesizer by name.')
-                synthesizer = baselines[synthesizer]
+            LOGGER.info('Trying to import synthesizer by name.')
+            synthesizer_name = synthesizer
+            if synthesizer in st_sdv_synthesizers + mt_sdv_synthesizers:
+                modality = 'single_table' if synthesizer in st_sdv_synthesizers else 'multi_table'
+                instance = BaselineSDVSynthesizer(synthesizer, modality)
```
amontanez24: Instead of directly returning an instance of BaselineSDVSynthesizer, can you add a function, `create_sdv_synthesizer_class` and in it return something like:
```python
    CustomSynthesizer = type(
        class_name, # Should be the synthesizer name in sdv
        (BaselineSynthesizer,),
        {
            '__module__': __name__,
            'get_trained_synthesizer': get_trained_synthesizer,
            'sample_from_synthesizer': sample_from_synthesizer,
        },
    )
```

pvk-developer (approved):

Issue #484, opened 2025-11-03
When running a benchmark locally, the `additional_datasets_folder` path should be the root path


npatki:
### Environment Details
* SDGym version: 0.11.0
* Python version: Any
* Operating System: Any

### Error Description
When running a benchmark, it is possible to provide an `additional_datasets_folder` parameter for providing custom datasets. The expected structure of this folder is shown [in the docs](https://docs.sdv.dev/sdgym/customization/datasets/dataset-format#folder-format).

Essentially, there is a root folder that contains sub-folders for single-table, multi-table, and sequential datasets. Inside each one, there is a prescribed structure for the data.

```
root_folder/ (local or S3 bucket)
|--- single_table/
     |--- dataset1/
          |--- ...
     |--- dataset2/
          |--- ...
|--- multi_table/
     |--- dataset3/
          |--- ...
|--- sequential/
     |--- ...
```

The problem is that there is an inconsistency in our 2 benchmarking methods:
1. When running an AWS benchmark, the user is expected to provide the root path (`additional_datasets_folder='root_folder/'`)
2. But when running a local benchmark, the user is expected to provide the path to the single-table subdirectory (`additional_datasets_folder='root_folder/single_table/'`)

### Expected Behavior
The expected input is to provide the path to the _root folder_, which is what we're currently doing for the AWS benchmark (1). We should update the local benchmark function to match this and be consistent.

PR #492, opened 2025-11-13
Incorporate the get_available_datasets functionality into the DatasetExplorer

### LINKED ISSUE #473: Incorporate the `get_available_datasets` functionality into the `DatasetExplorer`
### Problem Description
SDGym version 0.11.0 offers a `DatasetExplorer` class that you can use to view and explore all the datasets that are within a bucket. This class has enhanced functionality that summarizes all the datasets.

At the same time, there is a legacy method called [get_available_datasets](https://docs.sdv.dev/sdgym/customization/datasets/public-sdv-datasets) that simply lists all the datasets and some basic facts about them (their size and number of tables). Much of this functionality is already covered by `DatasetExplorer.summarize_datasets`.

However, the `get_avialable_datasets` functionality is a bit more lightweight. The reason is that `summarize_datasets` requires SDGym to actually load the data and perform some computations on the data -- which can take time. In the meantime, this function is more lightweight because it only needs to read information that is already available in the metainfo.yaml file.

### Expected behavior

Incorporate the functionality from `get_available_datasets` into `DatasetExplorer.list_datasets`. 

Inputs:
- (required) `modality`" The modality of the data to summarize. This should be one of 'single_table', 'multi_table' or 'sequential'.
- `output_filepath`: A string with the full output filepath where the results will be written. This should end in .csv

Output: This function produces the same dataframe as `get_available_datasets` does today.  If the `output_filepath` is provided, then the same dataframe would be written as a CSV in the file provided.

```python

de = DatasetExplorer(
  s3_url='s3://my_bucket/', # optional
  aws_access_key_id='my_access_key', # optional
  aws_secret_access_key='my_secret' # optional
)

dataset_list = de.list_datasets(modality='single_table', output_filepath='my_dataset_list.csv')
```

After doing this, we can _remove the existing `get_available_datasets` function_, as it is no longer needed.

fealho:
CU-86b7b4wnj, Resolve #473

### DIFF
--- a/sdgym/result_explorer/result_handler.py
+++ b/sdgym/result_explorer/result_handler.py
@@ -16,6 +16,7 @@ RESULTS_FOLDER_PREFIX = 'SDGym_results_'
 metainfo_PREFIX = 'metainfo'
 RESULTS_FILE_PREFIX = 'results'
 NUM_DIGITS_DATE = 10
+REGEX_SYNTHESIZER_NAME = r'\s*\(\d+\)\s*$'
 
 
 class ResultsHandler(ABC):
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b7b4wnj SDGym - When running a benchmark locally, the `additional_datasets_folder` path should be the root path #484](https://app.clickup.com/t/86b7b4wnj)

R-Palazzo (commented):
Looks good!
Could we add a unit test for `list_datasets()`?

PR #493, opened 2025-11-14
When running a benchmark locally, the additional_datasets_folder path should be the root path

### LINKED ISSUE #484: When running a benchmark locally, the `additional_datasets_folder` path should be the root path

fealho:
CU-86b7b4wnj, Resolve #484

### DIFF
--- a/sdgym/__init__.py
+++ b/sdgym/__init__.py
@@ -12,12 +12,16 @@ __version__ = '0.11.2.dev0'
 
 import logging
 
-from sdgym.benchmark import benchmark_single_table
+from sdgym.benchmark import benchmark_single_table, benchmark_single_table_aws
 from sdgym.cli.collect import collect_results
 from sdgym.cli.summary import make_summary_spreadsheet
 from sdgym.dataset_explorer import DatasetExplorer
 from sdgym.datasets import get_available_datasets, load_dataset
-from sdgym.synthesizers import create_sdv_synthesizer_variant, create_single_table_synthesizer
+from sdgym.synthesizers import (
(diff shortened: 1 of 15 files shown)


sdv-team:
Task linked: [CU-86b7b4wnj SDGym - When running a benchmark locally, the `additional_datasets_folder` path should be the root path #484](https://app.clickup.com/t/86b7b4wnj)

PR #498, opened 2025-11-17
Make DatasetExplorer warn If there are no datasets in the bucket

### LINKED ISSUE #475: If there are no datasets in the bucket, the `DatasetExplorer` should show a warning and return an empty table
### Problem Description
SDGym v0.11.0 has a DatasetExplorer object that allows the user to summarize all the datasets in an S3 bucket (as long as they follow our data format). 

If a user-provided S3 bucket does not contain any datasets, then it returns an an empty table. 

### Expected behavior
If a user-provided S3 bucket does not contain any datasets, then:
- (a) Show the user a warning, and
- (b) Return a table that contains the column names (table header) but no rows in the table. The same table would be written into the `output_filepath`, if provided.

For example:
```python

de = DatasetExplorer(
  s3_url='s3://my_bucket/',
  aws_secret_access_key='my_secret')

de.summarize_datasets(modality='single_table')
Warning: The provided S3 URL 's3://my_bucket/'' does not contain any datasets of modality 'single_table'.
```
Image

fealho:
CU-86b7a4r62, Resolve #475.

### DIFF
--- a/sdgym/dataset_explorer.py
+++ b/sdgym/dataset_explorer.py
@@ -3,12 +3,12 @@
 import warnings
 from collections import defaultdict
 from pathlib import Path
-from urllib.parse import urlparse
 
 import pandas as pd
 from sdv.metadata import Metadata
 
 from sdgym.datasets import BUCKET, _get_available_datasets, _validate_modality, load_dataset
+from sdgym.s3 import _validate_s3_url
 
 SUMMARY_OUTPUT_COLUMNS = [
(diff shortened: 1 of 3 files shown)


sdv-team:
Task linked: [CU-86b7a4r62 SDGym - If there are no datasets in the bucket, the `DatasetExplorer` should show a warning and return an empty table #475](https://app.clickup.com/t/86b7a4r62)

PR #499, opened 2025-11-17
Add input validation for the DatasetExplorer class and functions

### LINKED ISSUE #474: Add input validation for the `DatasetExplorer` class and functions
### Problem Description
SDGym v0.11.0 has a `DatasetExplorer` object that allows the user to summarize all the datasets in an S3 bucket (as long as they follow our data format). This object is currently missing some input validation -- for both the init and the its functions.

### Expected behavior

**During init**: We should validate that the `s3_url` (if provided) is a valid S3 URL -- it should (a) start with `s3://` and (b) should not contain any unexpected characters/formatting.

```python

de = DatasetExplorer(
  s3_url='malformed_bucket_name//?',
  aws_secret_access_key='my_secret'
)

Error: The provided s3_url parameter ('malformed_bucket_name//?') is not a valid S3 URL. 
Please provide a string that starts with 's3://' and refers to a AWS bucket.
```

**When running other functions**: When running a function like `summarize_datasets`, we should validate that the `output_filepath` doesn't already exist -- we don't want the user to accidentally override the file.

```python
dataset_summary = de.summarize_datasets(
  modality='single_table',
  output_filepath='datasets_summary.csv'
)

Error: The file 'datasets_summary.csv' already exists. Please provide a new 'output_filepath'.
```

fealho:
CU-86b7a4r6a, Resolve #474

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,18 @@
 # History
 
+## v0.12.0 - 2025-11-20
+
+### New Features
+
+* Rename create_sdv_synthesizer_variant to create_synthesizer_variant - Issue [#491](https://github.com/sdv-dev/SDGym/issues/491) by @R-Palazzo
+* SDGym should be able to automatically discover SDV Enterprise synthesizers - Issue [#481](https://github.com/sdv-dev/SDGym/issues/481) by @R-Palazzo
+* Incorporate the `get_available_datasets` functionality into the `DatasetExplorer` - Issue [#473](https://github.com/sdv-dev/SDGym/issues/473) by @fealho
+
+### Bugs Fixed
+
+* Update result aggregation logic in the ResultExplorer to match new naming schema - Issue [#494](https://github.com/sdv-dev/SDGym/issues/494) by @R-Palazzo
+* When running a benchmark locally, the `additional_datasets_folder` path should be the root path - Issue [#484](https://github.com/sdv-dev/SDGym/issues/484) by @fealho
+
 ## v0.11.1 - 2025-11-03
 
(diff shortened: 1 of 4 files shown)


sdgym/dataset_explorer.py
```diff
@@ -26,11 +26,47 @@ class DatasetExplorer:
...
+            raise ValueError(error_msg)
+
+        bucket_name = parsed.netloc
+        # Only bucket-level URLs are allowed (no keys/paths), allow optional trailing slash
+        if not bucket_name or parsed.path not in ('', '/'):
+            raise ValueError(error_msg)
+        if parsed.params or parsed.query or parsed.fragment:
+            raise ValueError(error_msg)
+
+        return bucket_name
```
pvk-developer: This should go into `s3.py` and be a function.

sdgym/dataset_explorer.py
```diff
@@ -240,12 +276,19 @@ def _validate_output_filepath(self, output_filepath):
...
+        if output_filepath:
+            path_obj = Path(output_filepath)
+            if path_obj.suffix != '.csv':
+                raise ValueError(
+                    f"The 'output_filepath' has to be a .csv file, provided: '{output_filepath}'."
+                )
+            if path_obj.exists():
+                raise ValueError(
+                    f"The file '{path_obj}' already exists. Please provide a new 'output_filepath'."
+                )
```
pvk-developer: can we move this to a `validate_output_path` function since it is doing now multiple checks

pvk-developer (approved):

rwedge (approved):

PR #517, opened 2025-12-19
Define internal single and multi table methods to run on GCP

### LINKED ISSUE #515: Define internal single and multi table methods to run on GCP
### Problem Description

The SDGym monthly benchmark execution is being migrated from AWS to Google Cloud Platform (GCP) compute instances. To support this, a new internal helper method, _benchmark_multi_table_compute_gcp, should be introduced to run the multi-table SDGym benchmark on GCP.

Starting with this implementation, the GCP compute instances should also install sdv_enterprise, enabling the inclusion of enterprise synthesizers in benchmark runs.

GCP should be used only for computation. All benchmark artifacts and results must continue to be stored in AWS (S3), consistent with the current benchmark workflow.

### Expected behavior

- To avoid conflicts with the existing AWS benchmark logic, create a new internal _benchmark folder.
- Define _benchmark_multi_table_compute_gcp inside this folder.
- The method should used similar logic to what is done in `benchmark_multi_table_aws`
```python
def _benchmark_multi_table_compute_gcp(
    output_destination,
    credential_filepath,
    compute_config=None,
    synthesizers=DEFAULT_MULTI_TABLE_SYNTHESIZERS,
    sdv_datasets=DEFAULT_MULTI_TABLE_DATASETS,
    additional_datasets_folder=None,
    limit_dataset_size=False,
    compute_quality_score=True,
    compute_diagnostic_score=True,
    sdmetrics=None,
    timeout=None,
):
````
### Parameters
- `output_destination`: An S3 bucket destination, matching the behavior of benchmark_multi_table_aws.
- `credential_filepath`: Path to a JSON file containing all required credentials. Expected structure:
```json
{
    "aws": {
        "aws_access_key_id": "<key_id>",
        "aws_secret_access_key": "<access_key_id>"
    },
    "gcp": {
        "type": "value",
        "project_id": "value",
        "private_key_id": "value",
        "private_key": "value",
        "client_email": "value",
        "client_id": "value",
        "auth_uri": "value",
        "token_uri": "value",
        "auth_provider_x509_cert_url": "value",
        "client_x509_cert_url": "value",
        "universe_domain": "value",
        "gcp_project": "value",
        "gcp_zone": "value"
    },
    "sdv": {
        "username": "value",
        "license_key": "value"
    }
}

```
- `compute_config`
A dictionary defining all parameters required to launch and manage compute instances.
- For now, this may default to None, in which case an internal default configuration is used.
- In the future, this may be exposed to users in a manner similar to AWS.
- Default configuration example:
```python
{
    "gcp": {
        "name_prefix": "sdgym-run",
        "machine_type": "n1-standard-8",
        "source_image": (
            "projects/deeplearning-platform-release/global/images/family/"
            "common-cu128-ubuntu-2204-nvidia-570"
        ),
        "gpu_type": "nvidia-tesla-t4",
        "gpu_count": 1,
        "install_nvidia_driver": False,
        "delete_on_success": True,
        "delete_on_error": True,
        "stop_fallback": True,
    },
    "aws": {
        "name_prefix": "sdgym-run",
        "ami": "ami-080e1f13689e07408",
        "instance_type": "g4dn.4xlarge",
        "volume_size_gb": 100,
    },
}

```
- Other parameters
   - All remaining parameters should mirror the behavior of `benchmark_multi_table_aws`.
### Additional context

This issue should define the full lifecycle required to launch, run, and terminate a GCP compute instance.
To remain consistent with the existing AWS benchmark workflow, the following requirements apply:
- Use a GPU-enabled GCP instance
   - Recommended GPU types:
   - nvidia-tesla-t4
   - nvidia-l4 (widely available and cost-effective)
- Ensure robust cleanup logic
   - Use a trap or equivalent mechanism so that the instance is stopped or deleted if:
       - the benchmark completes successfully, or
       - the benchmark fails due to an error
       - This prevents accidental billing from orphaned instances.
- At the end of this issue, a multi-table benchmark should be run on gcp with sdv_enterprise synthesizers and the results uploaded to aws
- Define the equivalent `single-table` method

R-Palazzo:
Resolve #515
CU-86b7w0f8w

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -70,7 +70,7 @@ sdgym = { main = 'sdgym.cli.__main__:main' }
 [project.optional-dependencies]
 dask = ['dask', 'distributed']
 realtabformer = [
-    'realtabformer>=0.2.3',
+    'realtabformer>=0.2.3,!=0.2.4',
     "torch>=2.6.0",
     'transformers<4.51',
 ]


sdgym/benchmark.py
```diff
@@ -1576,7 +1577,9 @@ def _store_job_args_in_s3(output_destination, job_args_list, s3_client):
     job_args_key = f'{path}{job_args_key}' if path else job_args_key
 
     serialized_data = cloudpickle.dumps(job_args_list)
-    s3_client.put_object(Bucket=bucket_name, Key=job_args_key, Body=serialized_data)
+    compressed = gzip.compress(serialized_data, compresslevel=1)
```
R-Palazzo: The `s3_client.put_object` is not designed for large objects, and the `job_arg_list` file is 500MB+ for multi-table dataset, so we compress it here.

sdgym/_benchmark/benchmark.py
```diff
@@ -0,0 +1,460 @@
...
+        aws configure set default.region '{S3_REGION}'
+
+        log "======== Create Virtual Environment =========="
+        python3 -m venv ~/env
+        source ~/env/bin/activate
+
+        log "======== Install Dependencies =========="
+        pip install --upgrade pip
+        pip install "{config['sdgym_install']}"
+
```
amontanez24: where is the optional install of sdv-enterprise?

sdgym/benchmark.py
amontanez24: this is good for now 👍 I think in the future we should push towards not uploading all of the arguments and instead passing them as strings to the cloud instance. The cloud instance can then download or create anything it needs to

sdgym/_benchmark/benchmark.py
```diff
@@ -0,0 +1,463 @@
...
+    if not output_destination.startswith('s3://'):
+        return ''
+
+    parsed = urlparse(output_destination)
+    bucket = parsed.netloc
+    prefix = parsed.path.lstrip('/').rstrip('/')
+    if prefix:
+        prefix = f'{prefix}/logs'
+    else:
+        prefix = 'logs'
```
pvk-developer:
```suggestion
    prefix = f'{prefix}/logs' if prefix else 'logs'
```

tests/unit/_benchmark/test_benchmark.py
```diff
@@ -0,0 +1,551 @@
...
+    # Assert
+    assert 'cleanup()' in script
+    assert 'dmesg | tail -50' in script
+    assert 'latest/meta-data/instance-id' in script
+    assert 'aws ec2 terminate-instances' in script
+    assert 'shutdown -h now' not in script
+    assert 'compute.googleapis.com' not in script
+
+
+def test_terminate_instance_gcp():
```
pvk-developer: Rename all the tests in this file that are with a single `_` to `test__`.

.github/workflows/minimum.yml
```diff
@@ -30,7 +30,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-          python -m pip install --upgrade pip
+          python -m pip install --upgrade "pip<24.2"
```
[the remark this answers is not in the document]
R-Palazzo: This is because `grpcio` wheels for macOS have incorrect wheel tags, and pip 24.2+ strictly validates wheel compatibility and crashes on an incorrect tag:
https://github.com/sdv-dev/SDGym/actions/runs/20426827536/job/58688803384

amontanez24 (approved):

PR #520, opened 2025-12-26
Remove deprecated parameters 

### LINKED ISSUE #519: Remove deprecated parameters
### Problem Description

We have a number of deprecated parameters in the [benchmark_single_table](https://github.com/sdv-dev/SDGym/blob/b8b7b133a5028c621c7dc5329ed4bbf32940aa3f/sdgym/benchmark.py#L1224) function that still remain in the code. There are a lot of code artifacts tied to these parameters that clutter the codebase and make it difficult to read. We should clean this code up by removing the parameters and any functions or conditionals that are related to them. The deprecated parameters are:
- `run_on_ec2` - this is not needed since we have a separate function for it now
- `output_filepath` - this was replaced by `output_destination`
- `detailed_results_folder` - Also replaced by `output_destination`
- `multi_processing_config` - We never actually use the multi-processing feature in runs.

### Expected behavior

- The parameters above should be completely removed from the function
- Any code that handles these parameters or exists because of them should be deleted

### Additional context

- The `multi_processing_config` parameter has a lot of code tied to it that we don't really use or need. This can be seen here 
https://github.com/sdv-dev/SDGym/blob/b8b7b133a5028c621c7dc5329ed4bbf32940aa3f/sdgym/benchmark.py#L835-L852
- We don't advertise it and it defaults to None, so in practice our runs just map the run to the job_arg_list directly
https://github.com/sdv-dev/SDGym/blob/b8b7b133a5028c621c7dc5329ed4bbf32940aa3f/sdgym/benchmark.py#L849
- This defeats the purpose of all these nested function. We could instead have the benchmark_single_table method just call the [_run_job](https://github.com/sdv-dev/SDGym/blob/b8b7b133a5028c621c7dc5329ed4bbf32940aa3f/sdgym/benchmark.py#L734) function directly.

fealho:
CU-86b7y0qa1, Resolve #519

### DIFF
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -1,6 +1,5 @@
 """Main SDGym benchmarking module."""
 
-import concurrent
 import logging
 import math
 import multiprocessing
@@ -42,7 +41,7 @@ from sdmetrics.single_table import DCRBaselineProtection
 from sdgym.datasets import get_dataset_paths, load_dataset
 from sdgym.errors import BenchmarkError, SDGymError
 from sdgym.metrics import get_metrics
-from sdgym.progress import TqdmLogger, progress
+from sdgym.progress import TqdmLogger
 from sdgym.result_writer import LocalResultsWriter, S3ResultsWriter
 from sdgym.s3 import (
     S3_PREFIX,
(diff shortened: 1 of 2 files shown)


sdv-team:
Task linked: [CU-86b7y0qa1 SDGym - Remove deprecated parameters #519](https://app.clickup.com/t/86b7y0qa1)

PR #526, opened 2026-01-08
Private S3 bucket access fails in benchmark_multi_table_aws despite valid credentials

### LINKED ISSUE #525: Private S3 bucket access fails in benchmark_multi_table_aws despite valid credentials
### Environment Details

* SDGym version: multi-table feature branch

### Error Description

When running benchmark_multi_table_aws with valid AWS credentials, the additional_datasets_folder parameter should allow access to private S3 buckets that the credentials have permission to read.

Currently, the benchmark fails with:

`ClientError: An error occurred (AccessDenied) when calling the ListObjectsV2 operation: Access Denied`

This happens even though the provided credentials have read access to the specified bucket.

### Steps to reproduce

```python
from sdgym import benchmark_multi_table_aws

access_key = 'TODO: Add valid credential that can access sdv-datasets-private'
secret_key = ''
suffix = 'template'
output_destination_aws = f's3://sdgym-benchmark/Debug/bug_hunt_multi_table_benchmark/{suffix}'

results = benchmark_multi_table_aws(
    output_destination=output_destination_aws,
    aws_access_key_id=access_key,
    aws_secret_access_key=secret_key,
    synthesizers=['HMASynthesizer'],
    sdv_datasets=['fake_hotels', 'Biodegradability'],
    additional_datasets_folder='s3://sdv-datasets-private',
)
```

### Expected behavior

The benchmark should be able to list and read datasets from the private S3 bucket specified in additional_datasets_folder when valid credentials are provided.

R-Palazzo:
Resolve #525
CU-86b82teqh

In `_generate_job_arg_list`, we were previously looking for the `aws_access_key` and `aws_access_key_id` environment variables. However, the method already receives an `s3_client` parameter with the correct credentials. I propagated its use and tried to rely on it whenever possible instead of the AWS key parameters. This way, the same s3_client is used consistently across the benchmark, let me know if it makes sense.

I tested this by running the code from the issue (an AWS benchmark using a private bucket). The results are available [here](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?prefix=Debug/bug_hunt_multi_table_benchmark/template/multi_table/SDGym_results_01_08_2026/&region=us-east-1). I stopped the instance to avoid running it for too long, but it successfully generated results for several private datasets.

### DIFF
--- a/.github/workflows/minimum.yml
+++ b/.github/workflows/minimum.yml
@@ -12,7 +12,7 @@ concurrency:
 jobs:
   minimum:
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 45
+    timeout-minutes: 60
     strategy:
       matrix:
         python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']


sdgym/datasets.py
```diff
@@ -242,8 +255,7 @@ def load_dataset(
     dataset,
     datasets_path=None,
     bucket=None,
-    aws_access_key_id=None,
-    aws_secret_access_key=None,
```
pvk-developer: This should still be valid option to pass to this public method (for usability), if you want to load a dataset with this method using aws credentials now you would have to do it in two steps, create a client then run this method.

Also  update the args in the docstring.

amontanez24 (commented):
Looks good! I like the idea of using a different method within _generate_job_args_list to load the dataset

pvk-developer (approved):

.github/workflows/minimum.yml
```diff
@@ -12,7 +12,7 @@ concurrency:
 jobs:
   minimum:
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 45
+    timeout-minutes: 60
```
R-Palazzo: Since we install grcpio from source, the workflow often takes longer, sometimes over 45 minutes: 
https://github.com/sdv-dev/SDGym/actions/runs/20996176647/job/60353452983.

I increased the timeout to ensure the tests pass, but let me know if we should file a specific issue about the grpcio installation in the minimum workflow

PR #553, opened 2026-02-20
Update load_results to be able to filter on dataset or synthesizer

### LINKED ISSUE #551: Define `get_dataset_results()` and `get_model_results()` for the ResultExplorer
### Problem Description

The ResultExplorer class is designed to facilitate navigation and analysis of benchmark results and generated artifacts.
To extend its usability, we would like to introduce two methods that provide filtered, high-level summaries of benchmark results:
- `Resultexplorer.get_dataset_results(dataset_name)`
- `Resultexplorer.get_synthesizer_results(synthesizer_name)`

These methods should provide a concise view of the time and scores scoped either by dataset or by synthesizer.

### Expected behavior

- Define get_dataset_results(dataset_name)
- Define get_synthesizer_results(synthesizer_name)
- Both should look for the benchmark results and filter to keep only the following columns:
    - `Synthesizer/Dataset` for the dataset/synthesizer method respectively
    - `Adjusted_Total_Time`
    - `Adjusted_Quality_Score`
    - `Diagnostic_Score`
    - `Win`
- They must return a pandas dataframe  

### Additional context
- We should maintain consistency with our current practice when dealing with multiple ‘run/results.csv’ files for a given day. Specifically, if a synthesizer has been executed twice for the same dataset, then the output of get_dataset_results should include two separate rows.

Visually the output should look like this:
- `get_dataset_results()`

Image

- `get_synthesizer_results()`
Image

R-Palazzo:
Resolve #551
CU-86b8j5dp9

### DIFF
--- a/tests/integration/result_explorer/test_result_explorer.py
+++ b/tests/integration/result_explorer/test_result_explorer.py
@@ -65,20 +65,21 @@ def test_end_to_end_local(tmp_path):
 
 
 @pytest.mark.parametrize(
-    'dataset_names, synthesizer_names, summary',
+    'dataset_names, synthesizer_names, summary, expected_columns',
     [
         (
             ['fake_hotels'],
             ['HMASynthesizer'],
             True,
+            SUMMARY_COLUMNS,
         ),
-        (['fake_hotels'], None, False),
-        (None, ['HMASynthesizer'], False),
-        (None, None, False),
-        (None, None, True),
+        (['fake_hotels'], None, False, None),
+        (None, ['HMASynthesizer'], False, None),
+        (None, None, False, None),
+        (None, None, True, SUMMARY_COLUMNS),
     ],
 )
-def test_load_results_with_filters(dataset_names, synthesizer_names, summary):
+def test_load_results_with_filters(dataset_names, synthesizer_names, summary, expected_columns):
     """Test loading results with dataset and synthesizer and summary filters."""
     # Setup
     output_destination = 'tests/integration/result_explorer/_benchmark_results/'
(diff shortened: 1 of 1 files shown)


sdv-team:
Task linked: [CU-86b8j5dp9 SDGym - Define `get_dataset_results()` and `get_model_results()` for the ResultExplorer #551](https://app.clickup.com/t/86b8j5dp9)

sdgym/result_explorer/result_handler.py
```diff
@@ -183,26 +191,81 @@ def summarize(self, folder_name):
 
         return summarized_table, folder_to_results[folder_name]
 
-    def load_results(self, results_folder_name):
+    def _validate_load_results_filters(self, dataset_names, synthesizer_names, summary):
+        if dataset_names is not None and not isinstance(dataset_names, list):
+            raise ValueError('`dataset_names` must be a list of strings or None.')
+
+        if synthesizer_names is not None and not isinstance(synthesizer_names, list):
+            raise ValueError('`synthesizer_names` must be a list of strings or None.')
```
fealho: Needs to validate elements are strings. You could use SDV's `_is_list_of_type` method (although that one is in sdv.cag._utils) or copy paste it here.

sdgym/result_explorer/result_handler.py
```diff
@@ -183,26 +191,81 @@ def summarize(self, folder_name):
...
+                warning_message = (
+                    f'No results found in folder "{results_folder_name}" '
+                    f'matching the specified filters:\n'
+                    f'{filters_text}'
+                )
+            else:
+                warning_message = f'No results found in folder "{results_folder_name}".'
+
+            warnings.warn(warning_message)
+            return pd.DataFrame()
```
fealho: You shouldn't return an empty dataframe, otherwise if a method calls something like `df['Dataset']` they would get a `KeyError`. You need to keep the column schema. I think you can just remove this line and let the code run, and lines 267/68 will handle it for you.
```

**Review by fealho — APPROVED** [review]

File: tests/integration/result_explorer/test_result_explorer.py [hunk 3]
```diff
@@ -62,6 +63,76 @@ def test_end_to_end_local(tmp_path):
...
+    if summary:
+        expected_columns = {
+            'Synthesizer',
+            'Dataset',
+            'Adjusted_Total_Time',
+            'Adjusted_Quality_Score',
+            'Diagnostic_Score',
+        }
+    else:
+        expected_columns = set(expected_results.columns)
```
amontanez24: you can just add this to the things being passed in parametrize. You don't need if statements

Issue #545, opened 2026-02-13
Move the current benchmark configs to yaml files


R-Palazzo:
### Problem Description

The SDGym benchmark launcher logic is mostly hardcoded in Python (datasets, synthesizer splits, timeout, output destination, etc.). This makes it harder to maintain, extend, and validate configurations up front before launching compute.

We want to move the benchmark configurations into YAML files and introduce a configuration object that can load, validate, and execute benchmarks from those files while keeping the current workflows unchanged.

### Expected behavior

- Create a folder named `_benchmark_launcher/`. Inside it:
   - Add a `base_benchmark.yaml` that defines configuration shared between single-table and multi-table benchmarks, including:
      - Benchmark method parameters (e.g., timeout, output_destination, and other benchmark kwargs)
      - Compute service selection (gcp or aws)
      - Where to locate all necessary credentials (as references such as environment variable names)
   - Add a `benchmark_single_table.yaml` that defines:
      -  The benchmark jobs and the `instance_jobs` (each entry represents one instance run, containing a list of synthesizers and a list of datasets)
   - Add a `benchmark_multi_table.yaml` that defines:
      - The benchmark jobs and the `instance_jobs`
  - Add a new `benchmark_config.py` that defines a `BenchmarkConfig` class responsible for:
     - Creating a valid and ready-to-run benchmark configuration from YAML
     - The class must have defined the following methods:
        - `load_from_dict()`
        - `load_from_yaml()`
        - `validate()`
        - `to_dict()`
        - `to_yaml() `
- Add a new `benchmark_launcher.py` that defines a `BenchmarkLauncher` class responsible for:
     - Executing a benchmark config and managing the benchmark.
     - The class must have defined the following methods:
        - `launch()`
        - `get_status()`
        - `terminate()`
     -  `get_status()` and `terminate()` definitions are out of the scope of this issue. They will be defined in another issue
 
 - Update `run_benchmark.py` to use the new configuration flow (load yaml/dict config → validate → launch), ensuring the workflow output and behavior remain unchanged. 
 
### Additional context

- The BenchmarkConfig/BenchmarkLauncher is similar to our Metadata/Synthesizer connection in SDV.
- Running the existing workflows should produce the same behavior as before.
- Credential handling must remain safe:
   - credentials should never be saved or accesible by the configuration object
   - the configuration should reference credentials through environment variables and/or a temporary credentials file path (created externally, e.g., by the workflow)
[...]

PR #558, opened 2026-02-26
Include SDV-Enterprise in single-table benchmarks

### LINKED ISSUE #549: Include SDV-Enterprise in single-table benchmarks
### Problem Description

The current monthly benchmark executes all public SDV synthesizers, along with selected third-party synthesizers such as RealTabFormer.

However, the SDV-Enterprise single-table synthesizers are not currently included in the benchmark runs.

We already support SDV-Enterprise synthesizers in the multi-table benchmark (e.g., HSASynthesizer, IndependentSynthesizer). We want to extend similar support to the single-table benchmark.

### Expected behavior

Add the following SDV-Enterprise single-table synthesizers to the internal benchmark pipeline:
- SegmentSynthesizer
- XGCSynthesizer
- BootstrapSynthesizer

### Additional context

Before enabling these synthesizers in the official monthly benchmark:
- Perform a dry run on all 9 single-table benchmark datasets.
- Collect runtime and performance metrics.
- Based on the results, define an appropriate GCP job-splitting strategy to achieve a reasonable runtime
- The DP synthesizers will be added in a separate issue

R-Palazzo:
Resolve #549

I tested it on the `child` dataset [here](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?region=us-east-1&prefix=Debug/Issue_549_2/single_table/SDGym_results_02_26_2026/&showversions=false). It's currently running on the 9 benchmark datasets. My guess is they won't take long, so we don't need to split their job, but I will keep an eye on it and make updates if necessary.

In this PR, I'm also setting up an optimized split for the single- and multi-table benchmarks that uses 31 EC2 instances.
I'm testing it to confirm that we can launch this many instances at the same time. With this setup, we should expect both results (single- and multi-table) in roughly 5 days, since we timeout after 4 days on some datasets. In the previous run, we had single-table results after 10 days and even longer for multi-table, so this would be a first improvement.

Additionally, we will soon update how benchmarks are set up (solving https://github.com/sdv-dev/SDGym/issues/545), so the private method like `_get_benchmark_setup` starts to make the transition, but in the end we will be using yaml files only.

### DIFF
--- a/tests/integration/test_benchmark.py
+++ b/tests/integration/test_benchmark.py
@@ -1030,7 +1030,7 @@ def test_benchmark_multi_table_error_during_fit(mock_augment_tables):
     # Run
     result = benchmark_multi_table(
         synthesizers=['HMASynthesizer', 'MultiTableUniformSynthesizer'],
-        sdv_datasets=['Student_loan', 'fake_hotels'],
+        sdv_datasets=['fake_hotels_extended', 'fake_hotels'],
     )
 
     # Assert


sdgym/run_benchmark/run_benchmark.py
```diff
@@ -135,12 +188,13 @@ def main():
     aws_secret_access_key = os.getenv('AWS_SECRET_ACCESS_KEY')
     date_str = datetime.now(timezone.utc).strftime('%Y-%m-%d')
     modality = args.modality
-    for synthesizer_group in MODALITY_TO_SETUP[modality]['synthesizers_split']:
-        MODALITY_TO_SETUP[modality]['method'](
+    benchmark_setup = _get_benchmark_setup(modality)
+    for job in benchmark_setup['job_split']:
```
fealho: Change `job` to `synthesizers, datasets` for clarity, so:
```python3
for synthesizers, datasets in benchmark_setup['job_split']:
```

sdgym/run_benchmark/run_benchmark.py
```diff
@@ -13,91 +13,144 @@
...
+    'alarm',
+    'census',
+    'child',
+    'covtype',
+    'expedia_hotel_logs',
+    'insurance',
+    'intrusion',
+    'news',
+]
+MULTI_TABLE_DATASETS = [
```
amontanez24: can you move these to a yaml?

fealho (approved):

PR #566, opened 2026-03-05
Move the current benchmark configs to yaml files

### LINKED ISSUE #545: Move the current benchmark configs to yaml files

R-Palazzo:
Resolve #545
CU-86b8h52bh

It's a pretty big PR, thanks in advance for your review :)

Currently the `BenchmarkLauncher` only works for GCP.
To make it work on other compute servie (e.g. AWS) we will have to define benchmark methods that have similar parameter as the gcp ones (with `credentials`/`compute_config`): 
https://github.com/sdv-dev/SDGym/blob/3c2a2483668511fd3afa382aa1046188742ee4fa/sdgym/_benchmark/benchmark.py#L350

### DIFF
--- a/sdgym/_benchmark_launcher/benchmark_multi_table.yaml
+++ b/sdgym/_benchmark_launcher/benchmark_multi_table.yaml
@@ -1,5 +1,5 @@
 modality: multi_table
-output_destination: 's3://sdgym-benchmark/Debug/Benchmark_Launcher/'
+output_destination: &output_destination 's3://sdgym-benchmark/Debug/Benchmark_Launcher/'
 datasets_multi_table: &datasets_multi_table
   - WebKP
   - DCG
--- a/sdgym/_benchmark_launcher/benchmark_single_table.yaml
+++ b/sdgym/_benchmark_launcher/benchmark_single_table.yaml
@@ -1,5 +1,5 @@
 modality: single_table
-output_destination: 's3://sdgym-benchmark/Debug/Benchmark_Launcher/'
+output_destination: &output_destination 's3://sdgym-benchmark/Debug/Benchmark_Launcher/'
 datasets_single_table: &datasets_single_table
   - adult
   - alarm


sdgym/_benchmark_launcher/benchmark_config.py
```diff
@@ -0,0 +1,147 @@
...
+        """Return a python ``dict`` representation of the ``BenchmarkConfig``."""
+        config = {}
+        for key in CONFIG_KEYS:
+            value = getattr(self, f'{key}', None)
+            if value is not None:
+                config[key] = value
+
+        return deepcopy(config)
+
+    def __repr__(self):
```
pvk-developer:
```suggestion
    def __str__(self):
```

sdgym/_benchmark_launcher/_validation.py
```diff
@@ -0,0 +1,405 @@
...
+        and parameter.kind
+        in (inspect.Parameter.POSITIONAL_OR_KEYWORD, inspect.Parameter.KEYWORD_ONLY)
+    }
+    required_from_yaml = required - _INJECTED_PARAMS
+    missing = required_from_yaml - set(method_params)
+    if missing:
+        errors.append(
+            f'method_params: missing required parameters for {method_to_run.__name__}:'
+            f' {sorted(missing)}'
+        )
```
amontanez24: is this necessary? Can't we just have defaults for whatever they don't provide? The only one that I can see as required is the output destination, and I don't think that one should be grouped with the other parameters.

PR #591, opened 2026-04-09
Fix win computation when uploading benchmark result

### LINKED ISSUE #590: Fix win computation when uploading benchmark result
### Environment Details

* SDGym version: 0.14.2

### Error Description

Currently the upload benchmark result [workflow](https://github.com/sdv-dev/SDGym/actions/runs/24123924062/job/70383942210) fail with this error:
```python
KeyError: "['Adjusted_Quality_Score'] not in index"
````

This happens because older benchmark result tables do not include the adjusted score columns used during win computation.
To preserve compatibility with previous benchmark runs while keeping the current win computation logic consistent across all results, the workflow should handle missing adjusted columns gracefully instead of failing.

### Expected Behavior

The wins table computation should work for both old and new benchmark result tables.

- If the adjusted columns are missing, they should be added temporarily for the wins computation.
- When the UniformSynthesizer result is missing for a given run, the adjusted values should be computed using the following fallback rules:
   - Adjusted Total Time = Train Time + Sample Time
   - Adjusted Quality Score = Quality Score

- After the wins table is computed, any temporarily added adjusted columns should be removed so old result tables remain unchanged.

R-Palazzo:
Resolve #590
CU-86b9ax872

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,7 +84,7 @@ sdgym = { main = 'sdgym.cli.__main__:main' }
 dask = ['dask', 'distributed']
 realtabformer = [
     'realtabformer>=0.2.3,!=0.2.4',
-    "torch>=2.6.0",
+    "torch>=2.6.0,<2.11.0",
     'transformers<4.51',
 ]
 test = [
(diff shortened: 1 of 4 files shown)


sdv-team:
Task linked: [CU-86b9ax872 SDGym - Fix win computation when uploading benchmark result #590](https://app.clickup.com/t/86b9ax872)

fealho (approved):
Could you add a test case where `missing_adjusted_columns` is not empty?

frances-h (approved):

PR #592, opened 2026-04-10
Make BenchmarkConfig.compute support instance configuration in YAML

R-Palazzo:
Resolve #586
CU-86b94u5fd


sdv-team:
Task linked: [CU-86b94u5fd SDGym - Make BenchmarkConfig.compute support instance configuration in YAML #586](https://app.clickup.com/t/86b94u5fd)

sdgym/_benchmark_launcher/benchmark_base.yaml
```diff
@@ -4,6 +4,13 @@ method_params:
...
 compute:
-  service: 'gcp'
+  service: gcp
+  instance_type: n1-highmem-16
+  boot_image: projects/deeplearning-platform-release/global/images/family/common-cu128-ubuntu-2204-nvidia-570
+  root_disk_gb: 300
+  gpu_type: nvidia-tesla-t4
+  gpu_count: 1
+  swap_gb: 64
+  name_prefix: sdgym-run
```
R-Palazzo, line 14: I went with this approach to stay consistent with what already exists.

I wonder whether in the future it would be better to support a higher-level `image config` that resolves the image for the user, keep `boot_image` only as an explicit override, or support both.

```yaml
image:
    flavor: dlvm
    os: ubuntu
    os_version: "22.04"
    cuda: "12.8"
```

sdgym/_benchmark_launcher/utils.py
```diff
@@ -58,6 +58,35 @@
...
 _GCP_SERVICE_ACCOUNT_JSON_FILEPATH = 'GCP_SERVICE_ACCOUNT_JSON_FILEPATH'
 
+_REQUIRED_CANONICAL_KEYS = (
+    'instance_type',
+    'boot_image',
+    'root_disk_gb',
+)
+
+
+def _validate_required_keys(config, required_keys, config_name):
```
fealho: This is not called anywhere.

sdgym/_benchmark_launcher/utils.py
```diff
@@ -58,6 +58,35 @@
...
+    missing = [key for key in required_keys if key not in config or config[key] is None]
+    if missing:
+        missing_str = ', '.join(sorted(missing))
+        raise ValueError(f'Missing required keys in {config_name}: {missing_str}')
+
+
+def _resolve_compute_gcp(config):
+    """Convert canonical compute config keys to GCP-specific keys."""
+    result = dict(config)
+    result['machine_type'] = result.pop('instance_type')
```
fealho, line 71: Without a default, the `pop` will raise if the keys are missing. Maybe the `_validate_required_keys` was meant to be used here?

sdgym/_benchmark_launcher/_validation.py
```diff
@@ -61,12 +62,33 @@ def _validate_structure(config):
...
+    if gpu_count > 0 and not compute.get('gpu_type'):
+        errors.append('compute.gpu_type is required when compute.gpu_count > 0.')
+
+    return sorted(errors)
+
+
+def _validate_compute(compute):
+    service = compute.get('service')
+    if service != 'gcp':
+        return [f"compute.service: must be 'gcp'. Found: {service!r}"]
```
fealho: I think this is dead code. This method is only called [here](https://github.com/sdv-dev/SDGym/pull/592/changes#diff-f52411384331ecf2796bb9ccfe7c4796ec0356064ac69cf3591478b062c3c89cR79), but if `service != 'gcp'` it will fail [before](https://github.com/sdv-dev/SDGym/pull/592/changes#r3065435400) reaching that line, because `_METHODS` is only defined for gcp.

fealho (changes requested):

sdgym/_benchmark_launcher/benchmark_config.py
fealho, line 71: This is where I think it would fail that I was referencing in my other comment.

sdgym/_benchmark_launcher/utils.py
R-Palazzo: Good catch thanks, I forgot to remove it. Done in [4aa044b](https://github.com/sdv-dev/SDGym/pull/592/commits/4aa044b4b5c61f9e6ed5d31dc33d6addd7451b54)

sdgym/_benchmark_launcher/utils.py
R-Palazzo: No the validation is done beforehand. The resolve methods used by the BenchmarkLauncher expect the BenchmarkConfig to have already been validated, so there should not be missing keys here.

fealho (approved):

sdgym/_benchmark_launcher/utils.py
```diff
@@ -58,6 +58,27 @@
...
+    result['source_image'] = result.pop('boot_image')
+    result['disk_size_gb'] = result.pop('root_disk_gb')
+
+    return result
+
+
+def resolve_compute(compute):
+    """Resolve the compute configuration based on the benchmark config."""
+    return _resolve_compute_gcp(compute)
```
amontanez24: can we add a check here that determines the service we are using for compute and then do
```python
if service == 'gcp':
    return _resolve_compute_gcp()
elif service == 'aws:
    ...
else:
    raise Error('service must be one of ...')
```

amontanez24 (commented):
Just one comment!

sdgym/_benchmark_launcher/utils.py
R-Palazzo: Yes, done in [ad42d49](https://github.com/sdv-dev/SDGym/pull/592/commits/ad42d49d9718f89d9ea1268c6397b8e2d1cdc579)

amontanez24 (approved):

PR #594, opened 2026-04-14
Update the upload benchmark workflow to work with the BenchmarkLauncher

R-Palazzo:
Resolve #584
CU-86b92pcm6

### DIFF
--- a/sdgym/_benchmark_launcher/_storage_manager.py
+++ b/sdgym/_benchmark_launcher/_storage_manager.py
@@ -1,6 +1,7 @@
 from sdgym._benchmark_launcher.utils import resolve_credentials
 from sdgym.s3 import _list_s3_bucket_contents, get_s3_client, is_s3_path, parse_s3_path
-
+import pandas as pd
+import io
 
 def _validate_s3_output_destinations(instance_jobs):
     """Validate that all output destinations are S3 paths."""
@@ -62,3 +63,29 @@ class S3StorageManager(BaseStorageManager):
     def get_existing_filenames(self, output_destination):
         """Return the existing filenames for the given destination."""
         return {obj['Key'] for obj in self.list_files(output_destination)}
+
+    def build_instance_result_key(self, jobs, result_filename):
+        artifact_key_prefix = jobs[0]['artifact_key_prefix']
+        return f'{artifact_key_prefix}/{result_filename}'
+
+    def file_exists(self, output_destination, key):
+        existing_keys = self.get_existing_filenames(output_destination)
+        return key in existing_keys
+
+    def read_csv(self, output_destination, key):
+        s3_client, bucket_name = self._get_s3_resources(output_destination)
+        response = s3_client.get_object(Bucket=bucket_name, Key=key)
+        return pd.read_csv(io.BytesIO(response['Body'].read()))
+
+    def write_csv(self, result, output_destination, key):
+        s3_client, bucket_name = self._get_s3_resources(output_destination)
+        body = result.to_csv(index=False).encode()
+        s3_client.put_object(Body=body, Bucket=bucket_name, Key=key)
+
+    def write_instance_result(self, result, instance_name, jobs, result_filename):
+        if not jobs:
+            return
+
+        output_destination = jobs[0]['output_destination']
+        key = self.build_instance_result_key(jobs, result_filename)
+        self.write_csv(result, output_destination, key)
--- a/sdgym/_benchmark_launcher/_storage_manager.py
+++ b/sdgym/_benchmark_launcher/_storage_manager.py
@@ -1,7 +1,11 @@
-from sdgym._benchmark_launcher.utils import resolve_credentials
-from sdgym.s3 import _list_s3_bucket_contents, get_s3_client, is_s3_path, parse_s3_path
-import pandas as pd
 import io
+
+import pandas as pd
+
+from sdgym._benchmark_launcher.utils import resolve_credentials
+from sdgym.result_writer import S3ResultsWriter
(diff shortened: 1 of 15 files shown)


sdv-team:
Task linked: [CU-86b92pcm6 SDGym - Update the upload benchmark workflow to work with the BenchmarkLauncher #584](https://app.clickup.com/t/86b92pcm6)

sdgym/_benchmark_launcher/_storage_manager.py
```diff
@@ -62,3 +131,42 @@ def list_files(self, output_destination):
...
+        """Read a CSV artifact from S3."""
+        s3_client, bucket_name = self._get_s3_resources(output_destination)
+        response = s3_client.get_object(Bucket=bucket_name, Key=filename)
+        return pd.read_csv(io.BytesIO(response['Body'].read()))
+
+    def write_csv(self, result, output_destination, filename):
+        """Write a CSV artifact to S3."""
+        bucket_name, _ = parse_s3_path(output_destination)
+        file_path = f's3://{bucket_name}/{filename}'
+        self._get_writer().write_dataframe(result, file_path, index=False)
```
pvk-developer: Should we add description docstring about what the args should be.

sdgym/_benchmark_launcher/_storage_manager.py
```diff
@@ -62,3 +131,42 @@ def list_files(self, output_destination):
...
+
+    def update_metainfo(self, output_destination, filename, content):
+        """Update metainfo for an artifact."""
+        file_path = _build_s3_uri(output_destination, filename)
+        self._get_writer().write_yaml(data=content, file_path=file_path, append=True)
+
+    def delete(self, output_destination, key):
+        """Delete an artifact from storage."""
+        s3_client, bucket_name = self._get_s3_resources(output_destination)
+        s3_client.delete_object(Bucket=bucket_name, Key=key)
```
pvk-developer: Should this be in a try except if something is wrong, like the object is not found in S3 etc... And also log if something goes wrong or successful. That way we can debug what happened.

sdgym/run_benchmark/upload_benchmark_results.py
```diff
@@ -489,12 +490,27 @@ def get_result_explorer(
...
     if not result_explorer.all_runs_complete(folder_name):
-        LOGGER.warning(f'Run {folder_name} is not complete yet. Exiting.')
-        if github_env:
-            with open(github_env, 'a') as env_file:
-                env_file.write('SKIP_UPLOAD=true\n')
-
-        sys.exit(0)
+        launcher = result_explorer._handler.load_synthesizer(
+            f'{folder_name}/{KEY_BENCHMARK_LAUNCHER}'
+        )
```
amontanez24: it's weird to me that load_synthesizer returns a launcher. Can we make a different function for loading an artifact based on the key?

R-Palazzo:
@pvk-developer @amontanez24 I ran an end-to-end test with 4 instances and a 20-minute timeout. For this test, I removed the 1-day grace period. The results look like [this](https://docs.google.com/spreadsheets/d/1RpjZ8bNjXyMIpePqgNtsF7ZHhR0DNHb-/edit?usp=sharing&ouid=109041959931122073350&rtpof=true&sd=true). For a deeper look, all the artifacts are stored in AWS [here](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?prefix=Debug%2FIssue_584%2F&region=us-east-1&tab=objects).

sdgym/_benchmark_launcher/_storage_manager.py
R-Palazzo: I reorganized the `storage_manager` in [ef2bfa8](https://github.com/sdv-dev/SDGym/pull/594/commits/ef2bfa8cb64ca1ac4fda74d8469c05742e679409) so it always expects a `filepath`, which makes more sense I think

sdgym/_benchmark_launcher/benchmark_launcher.py
```diff
@@ -402,6 +437,79 @@ def get_job_status(self, dataset_names=None, synthesizer_names=None, instance_na
...
+            result_df = self._build_or_load_instance_results(instance_name)
+            self._storage_manager.write_csv(result=result_df, filepath=result_filepath)
+            self._storage_manager.delete(job_arg_filepath)
+            self._update_instance_metainfo(instance_name)
+
+        self.terminate(verbose=True)
+
+    def _save_from_storage_manager(self, filepath):
+        """Save the benchmark launcher to a file using the storage manager."""
+        self._storage_manager.save_pickle(self, filepath)
```
amontanez24: I think we can remove the underscore and call this save_to_file or save_to_cloud or something like that. I think the name should reflect the action rather than referencing the class that does the saving

sdgym/run_benchmark/upload_benchmark_results.py
```diff
@@ -489,12 +490,28 @@ def get_result_explorer(
...
+                with open(github_env, 'a') as env_file:
+                    env_file.write('SKIP_UPLOAD=true\n')
+
+            sys.exit(0)
+
+        LOGGER.warning(
+            f'Run {folder_name} is not complete yet, and the timeout has expired. '
+            'Remaining instances will be terminated.'
+        )
+        launcher._finalize()
```
amontanez24: you can remove the '_' from functions that are called meant to be called

amontanez24 (commented):
Looks good! I just have minor comments

amontanez24 (approved):

pvk-developer (approved):

PR #607, opened 2026-05-25
Update the benchmark to load datasets from multiple S3 buckets when None is specified

### LINKED ISSUE #604: Update the benchmark to load datasets from multiple S3 buckets when None is specified
### Problem Description

By default, the benchmark downloads data from the s3 bucket that we use for SDV demo datasets. In some cases, we also want the ability to load from multiple buckets as sources too. To provide this dynamic behavior, we can create an indexing function that tells us which datasets are available in each bucket, then we use the corresponding bucket to load the dataset.

### Expected behavior

Right now, if the bucket is not specified, it will default to the public bucket (see [line](https://github.com/sdv-dev/SDGym/blob/4eb841bf929171cccfa3f536cc62899fa009993c/sdgym/datasets.py#L136)). We want to update this to inspect our own buckets (public and private).
* Allow `bucket` to be a list of buckets.
* Validate that you can access the buckets with the given credentials. If any is inaccessible, raise an error.
* If bucket is a list, create with `get_available_demos` or `_genereate_dataset_info` to search based on modality.
* Load the dataset from a corresponding bucket.
* Do **not** save the dataset locally.

### Additional context

Right now, we only tackle the issue of unspecified buckets. In the future, we want to provide the ability for custom specification. Moreover, we want the ability to provide custom credentials for loading datasets.

R-Palazzo:
Resolve #604
CU-86b9zwpcu

This PR includes:
- Relying on SDV `download_demo()` to get `sdv_dataset`. This works for public datasets or for users with SDV-Enterprise installed. Otherwise, because on SDV there is some [validation](https://github.com/sdv-dev/SDV/blob/b5e2068de7ff41f9b0452738adc827ce3a587bf1/sdv/datasets/demo.py#L48) to allow access to the public bucket only, I reused the SDV logic here, so it works if the S3 client is created with valid credentials (from input keys or env variables). 
- Because we rely on `download_demo()`, we no longer save the dataset locally.
- Improvement in the `_generate_job_arg_list`: Now we're only passing `dataset_info` inside it. The `dataset_info` includes everything necessary to then download the data and metadata during execution.
- Added an integration test with a private dataset. Also tested it on AWS [here](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?region=us-east-1&prefix=Debug/Issue_604_AWS/&showversions=false) and GCP [here](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?region=us-east-1&prefix=Debug/Issue_604_GCP/&showversions=false).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,8 +64,7 @@ dependencies = [
     'XlsxWriter>=1.2.8',
     "rdt>=1.18.2;python_version<'3.14'",
     "rdt>=1.20.0;python_version>='3.14'",
-    "sdmetrics>=0.21.0;python_version<'3.14'",
-    "sdmetrics>=0.26.0;python_version>='3.14'",
+    "sdmetrics>=0.28.0",
     "sdv @ git+https://github.com/sdv-dev/SDV.git@main",
 ]
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,10 +64,8 @@ dependencies = [
     'XlsxWriter>=1.2.8',
     "rdt>=1.18.2;python_version<'3.14'",
     "rdt>=1.20.0;python_version>='3.14'",
-    "sdmetrics>=0.21.0;python_version<'3.14'",
-    "sdmetrics>=0.26.0;python_version>='3.14'",
-    "sdv>=1.21.0;python_version<'3.14'",
-    "sdv>=1.33.0;python_version>='3.14'",
+    "sdmetrics>=0.28.0",
+    "sdv @ git+https://github.com/sdv-dev/SDV.git@main",
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,8 +66,7 @@ dependencies = [
     "rdt>=1.20.0;python_version>='3.14'",
     "sdmetrics>=0.21.0;python_version<'3.14'",
     "sdmetrics>=0.26.0;python_version>='3.14'",
-    "sdv>=1.21.0;python_version<'3.14'",
-    "sdv>=1.33.0;python_version>='3.14'",
+    "sdv @ git+https://github.com/sdv-dev/SDV.git@main",
 ]
 
 [project.urls]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,8 +64,10 @@ dependencies = [
     'XlsxWriter>=1.2.8',
     "rdt>=1.18.2;python_version<'3.14'",
     "rdt>=1.20.0;python_version>='3.14'",
-    "sdmetrics>=0.28.0",
-    "sdv @ git+https://github.com/sdv-dev/SDV.git@main",
+    "sdmetrics>=0.21.0;python_version<'3.14'",
+    "sdmetrics>=0.26.0;python_version>='3.14'",
+    "sdv>=1.21.0;python_version<'3.14'",
+    "sdv>=1.33.0;python_version>='3.14'",
 ]
 
 [project.urls]
--- a/sdgym/benchmark.py
+++ b/sdgym/benchmark.py
@@ -40,7 +40,14 @@ from sdmetrics.reports.single_table import (
 )
 from sdmetrics.single_table import DCRBaselineProtection
 
-from sdgym.datasets import _load_dataset_with_client, get_dataset_paths
+from sdgym.datasets import (
+    SDV_DATASETS_PRIVATE_BUCKET,
+    SDV_DATASETS_PUBLIC_BUCKET,
+    _get_dataset_bucket_mapping,
+    _load_dataset_with_client,
+    _load_sdv_demo_dataset,
+    get_dataset_paths,
+)
 from sdgym.errors import BenchmarkError, SDGymError
 from sdgym.metrics import get_metrics
 from sdgym.progress import TqdmLogger
@@ -121,6 +128,14 @@ class JobArgs(NamedTuple):
     dataset_name: str
     modality: str
     output_directions: Optional[dict]
+
+
+class ResolvedDataset(NamedTuple):
+    """Resolved dataset data and metadata for benchmark job creation."""
+
+    name: str
+    data: Any
+    metadata: Any
 
 
 def _import_and_validate_synthesizers(synthesizers, custom_synthesizers, modality):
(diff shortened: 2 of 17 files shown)


sdv-team:
Task linked: [CU-86b9zwpcu SDGym - Update the benchmark to load datasets from multiple S3 buckets when None is specified #604](https://app.clickup.com/t/86b9zwpcu)

pyproject.toml
```diff
@@ -64,10 +64,8 @@ dependencies = [
     'XlsxWriter>=1.2.8',
     "rdt>=1.18.2;python_version<'3.14'",
     "rdt>=1.20.0;python_version>='3.14'",
-    "sdmetrics>=0.21.0;python_version<'3.14'",
-    "sdmetrics>=0.26.0;python_version>='3.14'",
-    "sdv>=1.21.0;python_version<'3.14'",
-    "sdv>=1.33.0;python_version>='3.14'",
+    "sdmetrics>=0.28.0",
+    "sdv @ git+https://github.com/sdv-dev/SDV.git@main",
```
R-Palazzo, line 68: TODO: Update after SDV release

sdgym/datasets.py
```diff
@@ -251,6 +262,88 @@ def _get_available_datasets(
     return pd.DataFrame(datasets_info)
 
 
+def _get_dataset_bucket_mapping(modality, buckets, s3_client, skip_inaccessible=False):
+    """Map SDV demo dataset names to the bucket they should be loaded from."""
+    dataset_buckets = {}
```
amontanez24, line 267: nitpick: can we call dataset_to_bucket

R-Palazzo: Yes, done in [c349d70](https://github.com/sdv-dev/SDGym/pull/607/commits/c349d704ae83eae493066bcc112fe1e86a6f6cd3)

sdgym/benchmark.py
```diff
@@ -1,7 +1,6 @@
 """Main SDGym benchmarking module."""
 
 import functools
-import gzip
```
R-Palazzo, line 4: Since the dataset is no longer included in the job_arg_list, we don't need to compress it. Also, it now contains one job by default, and from my tests, it's roughly 1KB.

sdgym/benchmark.py
```diff
@@ -359,47 +359,111 @@ def _generate_job_args_list(
...
+            job_args_list.append(
+                JobArgs(
+                    synthesizer=synthesizer,
+                    dataset_info=dataset_info,
+                    metrics=sdmetrics,
+                    timeout=timeout,
+                    compute_quality_score=compute_quality_score,
+                    compute_diagnostic_score=compute_diagnostic_score,
+                    compute_privacy_score=compute_privacy_score,
+                    dataset_name=dataset_info.name,
```
frances-h: Out of curiosity, why have the `dataset_name` as a separate attribute if it's just a copy of `dataset_info.name`?

frances-h (approved):

sdgym/benchmark.py
R-Palazzo: Good catch thanks, I forgot to remove it, done in [b97fa0d](https://github.com/sdv-dev/SDGym/pull/607/commits/b97fa0d89ed2c7db69789c5d315941fa4d7ead62)

sdgym/benchmark.py
```diff
@@ -822,21 +884,52 @@ def _format_output(
     return scores
 
 
+def _get_s3_client_from_result_writer(result_writer):
+    if isinstance(result_writer, S3ResultsWriter):
+        return result_writer.s3_client
+
+    return None
```
sarahmish, line 891: Just want to note that in the future, we should setup the s3 client independently from using the same credentials as writing the results.

sdgym/datasets.py
```diff
@@ -251,6 +262,87 @@ def _get_available_datasets(
...
+            raise ValueError(
+                f"Bucket '{bucket}' is not accessible with the provided credentials."
+            ) from error
+
+        for dataset_name in available_datasets['dataset_name'].tolist():
+            existing_bucket = dataset_buckets.get(dataset_name)
+            # If a dataset is available in multiple buckets, prefer the private one.
+            if existing_bucket is None or bucket == SDV_DATASETS_PRIVATE_BUCKET:
+                dataset_buckets[dataset_name] = bucket
+
```
sarahmish, line 295: maybe add this to the log notifying if a dataset appeared in multiple locations and which one will you use.

sdgym/datasets.py
```diff
@@ -251,6 +262,87 @@ def _get_available_datasets(
...
+            modality=modality,
+            dataset_name=dataset_name,
+            s3_bucket_name=bucket_name,
+        )
+        metadata = metadata.to_dict()
+    except ValueError:
+        if bucket != SDV_DATASETS_PRIVATE_BUCKET:
+            raise
+
+        data, metadata = _load_private_sdv_demo_dataset(
```
sarahmish, line 329: not sure about this but I think that you can get the session's credentials and pass them directly to `download_demo` instead of repeating the logic of loading with a private bucket. Not a big issue, but it's better to use the [feature](https://docs.sdv.dev/sdv/multi-table-data/data-preparation/loading-data/demo-data) that's already available in SDV instead of repeating.

https://docs.aws.amazon.com/boto3/latest/reference/core/session.html

sdgym/datasets.py
```diff
@@ -251,6 +262,87 @@ def _get_available_datasets(
...
+        data, metadata = download_demo(
+            modality=modality,
+            dataset_name=dataset_name,
+            s3_bucket_name=bucket_name,
+        )
+        metadata = metadata.to_dict()
+    except ValueError:
+        if bucket != SDV_DATASETS_PRIVATE_BUCKET:
+            raise
+
```
sarahmish: Since we know that `SDV_DATASETS_PRIVATE_BUCKET` should use the private loading function, it would be more optimal to directly use it
```python3
if bucket == SDV_DATASETS_PRIVATE_BUCKET:
     data, metadata = _load_private_sdv_demo_dataset(..)
else:
     data, metadata = download_demo(..)
```

I would remove the try-catch block since you are raising the exception unless it is a private bucket.

sdgym/datasets.py
```diff
@@ -5,8 +5,12 @@
 from pathlib import Path
 
 import appdirs
+import botocore
 import numpy as np
 import pandas as pd
+from sdv.datasets.demo import (
+    download_demo,
+)
```
sarahmish, line 21:
```suggestion
from sdv.datasets.demo import download_demo
```

sdgym/datasets.py
```diff
@@ -251,6 +255,62 @@ def _get_available_datasets(
...
+                dataset_buckets[dataset_name] = bucket
+
+    return dataset_buckets
+
+
+def _load_sdv_demo_dataset(
+    modality,
+    dataset_name,
+    bucket,
+    s3_client=None,
```
sarahmish, line 322: this is not used?

R-Palazzo: Good catch thanks,  removed in [4bd6e85](https://github.com/sdv-dev/SDGym/pull/607/commits/4bd6e85a19caea7c7d9c606e11509358520ade4a)

sarahmish (approved):

PR #611, opened 2026-05-26
Update the benchmark to launch one instance per (dataset, synthesizer) pair

### LINKED ISSUE #605: Update the benchmark to launch one instance per (dataset, synthesizer) pair
### Problem Description

A job is a single run of a dataset using a particular synthesizer and then evaluating the performance. Right now, we run jobs in parallel on a single instance, this can create large computational needs and can crash during benchmark execution. To alleviate any issues that can be caused by dataset/synthesizer needs, we want to update the benchmark config to always run only one job on an instance. 

_Note that this approach is not meant for resource optimization._

### Expected behavior

For each job:
1. launch an instance
2. run the benchmark
3. upload the results
4. terminate the instance

This change can be done on the [configuration](https://github.com/sdv-dev/SDGym/blob/main/sdgym/_benchmark_launcher/benchmark_multi_table.yaml) level by the benchmark launcher. What's important here is that we want to have ease in adding new datasets and synthesizers to the configuration. Therefore, I would propose that we can write a script to create the config file.

R-Palazzo:
Resolve #605
86b9zz525

- Tested for single-table [here](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?region=us-east-1&prefix=Debug/Issue_605/single_table/SDGym_results_05_26_2026/&showversions=false), it successfully launched 81 instances (9 synthesizers and 9 datasets).
- Tested for multi-table [here](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?region=us-east-1&prefix=Debug/Issue_605/multi_table/&showversions=false), it created 162 instances (54 datasets, 3 synthesizers).
- We need to store the list of datasets and synthesizers to run the benchmarks on. I stored them in a Python file, but I can move them to YAML if we prefer. To include a new synthesizer/dataset, we would just need to update these lists.


fealho (changes requested):
This runs the UniformSynthesizer in every instance. So for example, for the 81 single table runs, it ran 81 times. Is this intended?

Also, do we want to change [this](https://github.com/sdv-dev/SDGym/blob/3eeb8ae5d53ac3b3a6648ea1d19b9e19226a213a/sdgym/_benchmark_launcher/script.py#L223)? I don't think it affects the monthly run, because the config returns [before reaching that line](https://github.com/sdv-dev/SDGym/blob/3eeb8ae5d53ac3b3a6648ea1d19b9e19226a213a/sdgym/_benchmark_launcher/script.py#L218), but if you pass datasets/synthesizers, it still creates only 1 instance.

And could you double check the CLI path still works? I think maybe it depended on the yaml files you deleted. Something like this:

```
python sdgym/_benchmark_launcher/script.py --modality single_table --output-destination s3://bucket/test --synthesizers CTGANSynthesizer
```

amontanez24 (commented):
This looks good. I just want to confirm something. Does the instance that runs the workflow to launch the jobs download all the datasets for all jobs, or only the one for the job it launches?

I understand we are doing one job = 1 synthesizer and dataset, but I want to make sure the instance that launches the job isn't downloading too much

R-Palazzo:
> This looks good. I just want to confirm something. Does the instance that runs the workflow to launch the jobs download all the datasets for all jobs, or only the one for the job it launches?
> 
> I understand we are doing one job = 1 synthesizer and dataset, but I want to make sure the instance that launches the job isn't downloading too much

@amontanez24, yes, that was the case. The instance that runs the workflow was downloading all the datasets. I tried adding the `rel-bench` dataset and the workflow crashed when trying to download the data:

https://github.com/sdv-dev/SDGym/actions/runs/26508750022/job/78067938428

To avoid this in this [PR](https://github.com/sdv-dev/SDGym/pull/607) I moved the dataset loading during execution so it's only the gcp instances that will load the data. I'm running some test and check that it works end-to-end with the `rel-bench` datasets: 
- [Github Action](https://github.com/sdv-dev/SDGym/actions/runs/26522069764)
- [Results on AWS](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?prefix=Debug%2FIssue_605_with_rel_bench%2Fmulti_table%2FSDGym_results_05_27_2026%2F&region=us-east-1&tab=objects)

R-Palazzo:
> This runs the UniformSynthesizer in every instance. So for example, for the 81 single table runs, it ran 81 times. Is this intended?
> 
> Also, do we want to change [this](https://github.com/sdv-dev/SDGym/blob/3eeb8ae5d53ac3b3a6648ea1d19b9e19226a213a/sdgym/_benchmark_launcher/script.py#L223)? I don't think it affects the monthly run, because the config returns [before reaching that line](https://github.com/sdv-dev/SDGym/blob/3eeb8ae5d53ac3b3a6648ea1d19b9e19226a213a/sdgym/_benchmark_launcher/script.py#L218), but if you pass datasets/synthesizers, it still creates only 1 instance.
> 
> And could you double check the CLI path still works? I think maybe it depended on the yaml files you deleted. Something like this:
> 
> ```
> python sdgym/_benchmark_launcher/script.py --modality single_table --output-destination s3://bucket/test --synthesizers CTGANSynthesizer
> ```

Hi @fealho thanks for the review!
- Yes, it is intended for the UniformSynthesizer. However, when we aggregate the results to generate the monthly leaderboard, only one UniformSynthesizer per dataset is retained and used to compute `Adjusted_Total_Time` and `Adjusted_Quality_Score`. So it’s consistent between all synthesizers and datasets.
- Good catch for the `scripty.py`, I updated so it's always one instance per job
- Tested the CLI [here](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?region=us-east-1&prefix=Debug/Issue_605_CLI/&showversions=false), running:
```python
python sdgym/_benchmark_launcher/script.py --modality single_table --output-destination s3://sdgym-benchmark/Debug/Issue_605_CLI --synthesizers GaussianCopulaSynthesizer
```

fealho (approved):

amontanez24 (approved):

sdgym/_benchmark_launcher/utils.py
```diff
@@ -57,6 +57,11 @@
             'rel-salt',
             'rel-stack',
             'rel-trial',
+            'instacart_marketbasket_ml',
+            'MovieLens',
+            'rossmann',
+            'Telstra',
+            'walmart',
```
R-Palazzo, line 64: I checked it with Kalyan this morning and we want to have all the private demo datasets included in the benchmark

sarahmish (approved):
I prefer if the list of synthesizers and datasets is moved to a separate file as you suggested to make the processes of adding / removing datasets and synthesizers easy. That could be another issue

PR #622, opened 2026-06-09
Avoid loading full datasets when generating benchmark dataset details

### LINKED ISSUE #609: Avoid loading full datasets when generating benchmark dataset details
### Problem Description

When uploading benchmark results, SDGym generates `dataset details` that include row counts. For large datasets, this currently requires loading the dataset data, which can be slow and may crash due to memory usage (this was observed with `rel-bench` datasets). 

### Expected behavior

SDGym should compute accurate row counts for dataset details without loading the full dataset into pandas or saving the dataset locally. This should make benchmark result uploads more reliable for large datasets while preserving the existing `Dataset_Details.xlsx` output.

### Additional context

This should be tested with rel-bench datasets.

R-Palazzo:
Resolve #609
86ba43c4v

Tested for the multi-table benchmark upload  workflow [here](https://github.com/sdv-dev/SDGym/actions/runs/27151290663)


tests/integration/test_dataset_explorer.py
```diff
@@ -54,3 +56,29 @@ def test_dataset_explorer_empty_bucket_warns_and_returns_header_only(modality, t
...
+    summaries = explorer._load_and_summarize_datasets(
+        modality='multi_table',
+        datasets=['rel-arxiv'],
+    )
+
+    # Assert
+    assert len(summaries) == 1
+    assert summaries[0]['Dataset'] == 'rel-arxiv'
+    assert set(summaries[0]) == set(SUMMARY_OUTPUT_COLUMNS)
+    assert summaries[0]['Total_Num_Rows'] > 0
```
gsheni: Could we put in an actually number here?

tests/integration/test_dataset_explorer.py
```diff
@@ -54,3 +56,29 @@ def test_dataset_explorer_empty_bucket_warns_and_returns_header_only(modality, t
...
+    explorer = DatasetExplorer(
+        s3_url=SDV_DATASETS_PRIVATE_BUCKET,
+        aws_access_key_id=os.getenv('AWS_ACCESS_KEY_ID'),
+        aws_secret_access_key=os.getenv('AWS_SECRET_ACCESS_KEY'),
+    )
+
+    # Run
+    summaries = explorer._load_and_summarize_datasets(
+        modality='multi_table',
+        datasets=['rel-arxiv'],
```
gsheni, line 77: Has this been tested with other private datasets? I believe `rel-amazon` is the largest dataset (2.9GB). 
We should make sure that it works with that dataset (not for the integration test but test it locally (if we can)).

R-Palazzo: Yes tested locally with `rel-amazon` and also with the GitHub workflow [here](https://github.com/sdv-dev/SDGym/actions/runs/27151290663)

gsheni (approved):

sdgym/dataset_explorer.py
```diff
@@ -212,6 +226,66 @@ def get_data_summary(data):
...
+        has_header = False
+        for row in reader:
+            if not row:
+                continue
+
+            if not has_header:
+                has_header = True
+                continue
+
+            count += 1
```
sarahmish: I think this can be optimized

```suggestion
    next(reader)  # skip header
    count = sum(1 for row in reader if row)
```

sarahmish: I think you can also count the number of new lines without a csv reader which would make it even faster
```python3
def _count_csv_rows(csv_file):
    text_file = io.TextIOWrapper(csv_file, encoding='utf-8-sig', newline='')
    count = sum(1 for line in text_file if line.strip()) - 1 # -1 for the header
    text_file.detach()
    return max(count, 0)
```

sarahmish (commented):
Do we want to calculate the number of cells in the dataset too? number of rows x number of columns?

R-Palazzo:
> Do we want to calculate the number of cells in the dataset too? number of rows x number of columns?

@sarahmish this is something we could add, maybe in a separate issue. For now in the `Dataset_Details` table there are the `Total_Num_Columns` and `Total_Num_Rows`

sdgym/dataset_explorer.py
R-Palazzo: I kept the csv reader otherwise the tests were failing:
https://github.com/sdv-dev/SDGym/actions/runs/27419559432/job/81041188188

For example, the following was treated as 2 lines
```
id,text
1,"first line
second line"
```

sarahmish (approved):

amontanez24 (approved):

PR #628, opened 2026-07-16
Add TabDDPM to SDGym single table synthesizers

sarahmish:
original implementation is https://github.com/yandex-research/tab-ddpm/tree/main/tab_ddpm

### DIFF
--- a/sdgym/synthesizers/__init__.py
+++ b/sdgym/synthesizers/__init__.py
@@ -8,6 +8,7 @@ from sdgym.synthesizers.generate import (
 from sdgym.synthesizers.identity import DataIdentity
 from sdgym.synthesizers.column import ColumnSynthesizer
 from sdgym.synthesizers.realtabformer import RealTabFormerSynthesizer
+from sdgym.synthesizers.tabddpm import TabDDPMSynthesizer
 from sdgym.synthesizers.uniform import UniformSynthesizer, MultiTableUniformSynthesizer
 from sdgym.synthesizers.utils import (
     get_available_single_table_synthesizers,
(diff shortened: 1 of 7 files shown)


R-Palazzo (commented):
This is looking good!

Do we expect this PR to support running `TabDDPMSynthesizer` in the single-table benchmark, including running it on AWS and GCP?

If so, could we add integration tests for `benchmark_single_table()`?

For AWS and GCP, I previously tested this manually and shared the result folder in the PR. For example, for #549, this is the test [folder](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?region=us-east-1&prefix=Debug/Issue_549/single_table/&showversions=false)

I can help for testing it on AWS and GCP

amontanez24 (commented):
We're expecting that they will never update their library right? That is why we're moving all the code into here?

sarahmish:
> We're expecting that they will never update their library right? That is why we're moving all the code into here?

Yes exactly, the last update made to the repo was 3 years ago. It's unmaintained code that if we'd like to benchmark then we should copy over the code.

amontanez24 (approved):

R-Palazzo (approved):

sarahmish:
Results of the benchmark test are provided in this [folder](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?region=us-east-1&prefix=Debug/Sarah_TabDDPM_GCP_all_datasets/single_table/SDGym_results_07_24_2026/&showversions=false/) @R-Palazzo @amontanez24

PR #630, opened 2026-07-22
Add ClavaDDPM to SDGym multi table synthesizers

sarahmish:
original implementation is https://github.com/weipang142857/ClavaDDPM

The important class to review is `ClavaDDPM` and `ClavaDDPMSynthesizer` as these are the wrappers I wrote. I've written other code pieces, if it's taken from the source code directly, I mention "from" in the docstrings.

### DIFF
--- a/sdgym/synthesizers/clavaddpm.py
+++ b/sdgym/synthesizers/clavaddpm.py
@@ -18,7 +18,6 @@ import pandas as pd
 import sklearn
 import torch
 from packaging.version import Version
-from pandas.tseries.api import guess_datetime_format
 from sdv.metadata import Metadata
 from sklearn.cluster import KMeans
 from sklearn.mixture import BayesianGaussianMixture, GaussianMixture
@@ -45,6 +44,12 @@ def guess_array_datetime_format(values, sample_size=100, dayfirst=False):
     Returns:
         str: Datetime format string, or None if it can not be guessed.
     """
+    pandas_version = Version(pd.__version__).release[:2]
+    if pandas_version >= (2, 2):
+        from pandas.tseries.api import guess_datetime_format  # pandas >= 2.2.0
+    else:
+        from pandas._libs.tslibs.parsing import guess_datetime_format  # pandas < 2.2.0
+
     series = pd.Series(values)
     series = series.dropna()
     if series.empty:
--- a/sdgym/synthesizers/clavaddpm.py
+++ b/sdgym/synthesizers/clavaddpm.py
@@ -26,6 +26,8 @@ from sklearn.preprocessing import LabelEncoder, MinMaxScaler, OneHotEncoder
 from sdgym.synthesizers.base import MultiTableBaselineSynthesizer
 from sdgym.synthesizers.tabddpm import CAT_MISSING_VALUE, TabDDPM, ohe_to_categories
 
+SKLEARN_VERSION = Version(sklearn.__version__)
+
 
 def get_group_data_dict(np_data, group_id_attrs=[0]):
     """Grouping dictionary from pipeline_utils.py."""
@@ -298,8 +300,7 @@ def pair_clustering_keep_id(
 
         cat_one_hot = np.empty((cat_converted.shape[0], 0))
         for col in range(cat_converted.shape[1]):
-            sklearn_version = Version(sklearn.__version__).release
-            if sklearn_version[:2] >= (1, 2):
+            if SKLEARN_VERSION.release[:2] >= (1, 2):
                 encoder = OneHotEncoder(sparse_output=False)
             else:
                 encoder = OneHotEncoder(sparse=False)
@@ -330,15 +331,19 @@ def pair_clustering_keep_id(
     child_group_lengths = np.array([len(group) for group in child_group_data], dtype=int)
     num_clusters = min(num_clusters, len(cluster_data))
 
+    init_param = 'k-means++'
+    if SKLEARN_VERSION.release[:2] < (1, 1):
+        init_param = 'k-means'
+
     if clustering_method == 'kmeans':
-        kmeans = KMeans(n_clusters=num_clusters, n_init='auto', init='k-means++', random_state=seed)
+        kmeans = KMeans(n_clusters=num_clusters, n_init='auto', init=init_param, random_state=seed)
         kmeans.fit(cluster_data)
         cluster_labels = kmeans.labels_
     elif clustering_method == 'both':
         gmm = GaussianMixture(
             n_components=num_clusters,
             covariance_type='diag',
-            init_params='k-means++',
+            init_params=init_param,
             tol=0.0001,
             random_state=seed,
         )
@@ -348,7 +353,7 @@ def pair_clustering_keep_id(
         gmm = BayesianGaussianMixture(
             n_components=num_clusters,
             covariance_type='diag',
-            init_params='k-means++',
+            init_params=init_param,
             tol=0.0001,
             random_state=seed,
         )
--- a/sdgym/synthesizers/clavaddpm.py
+++ b/sdgym/synthesizers/clavaddpm.py
@@ -15,7 +15,9 @@ from datetime import datetime, timedelta
 
 import numpy as np
 import pandas as pd
+import sklearn
 import torch
+from packaging.version import Version
 from sklearn.cluster import KMeans
 from sklearn.mixture import BayesianGaussianMixture, GaussianMixture
 from sklearn.neighbors import NearestNeighbors
@@ -296,7 +298,12 @@ def pair_clustering_keep_id(
 
         cat_one_hot = np.empty((cat_converted.shape[0], 0))
         for col in range(cat_converted.shape[1]):
-            encoder = OneHotEncoder(sparse_output=False)
+            sklearn_version = Version(sklearn.__version__).release
+            if sklearn_version[:2] >= (1, 2):
+                encoder = OneHotEncoder(sparse_output=False)
+            else:
+                encoder = OneHotEncoder(sparse=False)
+
             column = cat_converted[:, col].reshape(-1, 1)
             cat_one_hot = np.concatenate((cat_one_hot, encoder.fit_transform(column)), axis=1)
 
--- a/sdgym/synthesizers/clavaddpm.py
+++ b/sdgym/synthesizers/clavaddpm.py
@@ -678,6 +678,10 @@ class ClavaDDPM:
             Multi-parent matching controls for unique matching.
         no_matching (bool):
             Multi-parent matching controls for no matching.
+        max_categories (int):
+            Maximum number of distinct values kept for a categorical column.
+            Columns with more are reduced to this many values, randomly sampled
+            from the data.
         device (str or None):
             Whether to use ``'cuda'`` or ``'cpu'``. If None, auto-select is used.
         seed (int):
@@ -707,6 +711,7 @@ class ClavaDDPM:
         matching_batch_size=1000,
         unique_matching=True,
         no_matching=False,
+        max_categories=100,
         device=None,
         seed=0,
         verbose=False,
@@ -729,6 +734,7 @@ class ClavaDDPM:
         self.matching_batch_size = matching_batch_size
         self.unique_matching = unique_matching
         self.no_matching = no_matching
+        self.max_categories = max_categories
         self.device = device
         self.seed = seed
         self.verbose = verbose
@@ -824,11 +830,22 @@ class ClavaDDPM:
             if name in self._synthetic_pk:
                 df[self._primary_key[name]] = np.arange(len(df))
 
+            # cap the cardinality of the discrete columns.
+            for col in self._discrete_cols[name]:
+                uniques = df[col].dropna().unique()
+                if len(uniques) > self.max_categories:
+                    kept = np.random.choice(uniques, size=self.max_categories, replace=False)
+                    outside = ~df[col].isin(kept)
+                    df.loc[outside, col] = np.random.choice(kept, size=int(outside.sum()))
+
             date_info = {}
             for col, datetime_format in self._datetime_cols[name].items():
                 if datetime_format is None:
                     datetime_array = df[df.notna()].astype(str).to_numpy()
                     datetime_format = guess_array_datetime_format(datetime_array)
+                    df[col] = pd.to_datetime(
+                        df[col], format=datetime_format, errors='coerce'
+                    ).dt.strftime()
 
                 days_since, earliest = calculate_days_since_earliest_date(df[col], datetime_format)
                 df[col] = np.asarray(days_since, dtype=float)
--- a/sdgym/synthesizers/clavaddpm.py
+++ b/sdgym/synthesizers/clavaddpm.py
@@ -722,6 +722,17 @@ class ClavaDDPM:
             ))
             self._children[parent].append(child)
 
+        # Resolve a primary key for every table.
+        self._primary_key = {}
+        self._synthetic_pk = set()
+        for name, table_meta in meta['tables'].items():
+            pk = table_meta.get('primary_key')
+            if pk is None:
+                pk = f'__{name}_pk__'
+                self._synthetic_pk.add(name)
+
+            self._primary_key[name] = pk
+
         # Preprocessing
         self._discrete_cols = {}
         self._datetime_cols = {}  # name -> {col: datetime_format}
@@ -773,6 +784,9 @@ class ClavaDDPM:
         for name in self._table_names:
             output_cols = list(data[name].columns)
             df = data[name].copy().reset_index(drop=True)
+            # tables without a primary key get an internal row-id
+            if name in self._synthetic_pk:
+                df[self._primary_key[name]] = np.arange(len(df))
 
             date_info = {}
             for col, date_format in self._datetime_cols[name].items():
--- a/sdgym/synthesizers/clavaddpm.py
+++ b/sdgym/synthesizers/clavaddpm.py
@@ -10,7 +10,7 @@ https://github.com/weipang142857/ClavaDDPM/tree/main.
 import logging
 import random
 import sys
-from collections import defaultdict
+from collections import Counter, defaultdict
 from datetime import datetime, timedelta
 
 import numpy as np
@@ -18,7 +18,7 @@ import pandas as pd
 import sklearn
 import torch
 from packaging.version import Version
-from pandas.core.tools.datetimes import _guess_datetime_format_for_array
+from pandas.tseries.api import guess_datetime_format
 from sdv.metadata import Metadata
 from sklearn.cluster import KMeans
 from sklearn.mixture import BayesianGaussianMixture, GaussianMixture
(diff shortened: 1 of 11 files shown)


sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1048 @@
...
+        for name in self._table_names:
+            output_cols = list(data[name].columns)
+            df = data[name].copy().reset_index(drop=True)
+            # tables without a primary key get an internal row-id
+            if name in self._synthetic_pk:
+                df[self._primary_key[name]] = np.arange(len(df))
+
+            date_info = {}
+            for col, date_format in self._datetime_cols[name].items():
+                if date_format is None:
```
pvk-developer: Is the `date_format` important for precision ? If it is we could do a similar approach as the one used in RDT:
```python
from pandas.core.tools.datetimes import _guess_datetime_format_for_array

if datetime_format is None:
    datetime_array = df[df.notna()].astype(str).to_numpy()
    datetime_format = _guess_datetime_format_for_array(datetime_array)
```

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1048 @@
...
+                'parents': list(self._parents[name]),
+                'label_encoders': label_encoders,
+                'date_info': date_info,
+            }
+
+        self._clustering()
+        self._training()
+
+        self._fitted = True
+        return self
```
pvk-developer: Why we return `self` here?

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1048 @@
...
+    def _clustering(self):
+        """clava_clustering: cluster every parent->child relationship."""
+        relation_order_reversed = self._relation_order[::-1]
+        self._all_group_lengths_prob_dicts = {}
+
+        for parent, child in relation_order_reversed:
+            if parent is None:
+                continue
+            if self.verbose:
+                sys.stdout.write(f'Clustering {parent} -> {child}')
```
pvk-developer: This overlaps with the steps:
```bash
Clustering hotels -> guestsTraining None -> hotelsStep 500/1000 MLoss: 1.0474 GLoss: 1.0276 Sum: 2.075
Step 1000/1000 MLoss: 0.7815 GLoss: 0.9702 Sum: 1.7517
Training hotels -> guestsStep 500/1000 MLoss: 3.1458 GLoss: 1.0022 Sum: 4.148
```

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1048 @@
...
+            self._tables[parent]['df'] = parent_df
+            self._tables[child]['df'] = child_df
+            self._all_group_lengths_prob_dicts[(parent, child)] = group_lengths_prob_dicts
+
+    def _training(self):
+        """clava_training: train one TabDDPM per relationship (child_training)."""
+        self._models = {}
+        for parent, child in self._relation_order:
+            if self.verbose:
+                sys.stdout.write(f'Training {parent} -> {child}')
```
pvk-developer: Overlaps with the steps printed afterwards:
```bash

Clustering hotels -> guestsTraining None -> hotelsStep 500/1000 MLoss: 1.0474 GLoss: 1.0276 Sum: 2.075
Step 1000/1000 MLoss: 0.7815 GLoss: 0.9702 Sum: 1.7517
Training hotels -> guestsStep 500/1000 MLoss: 3.1458 GLoss: 1.0022 Sum: 4.148
````

File: sdgym/synthesizers/clavaddpm.py [hunk 5]
```diff
@@ -0,0 +1,1048 @@
...
+                than the original tables. Defaults to ``1.0``.
+        """
+        if not self._fitted:
+            raise RuntimeError('The synthesizer has not been fitted; call fit() first.')
+
+        synthetic_tables = {}
+
+        for parent, child in self._relation_order:
+            if self.verbose:
+                sys.stdout.write(f'Generating {parent} -> {child}')
```
pvk-developer: This prints `None` if the parent has no parent, also overlaps with `Sampled`
```bash
Generating None -> hotelsSampled 10/10 rows
Generating hotels -> guests
```

pvk-developer (commented):
When fitting I also see the steps (they come from `TabDDPM`), if `verbose` is set to `False` do we need to display them?

pvk-developer:
During fit there is a point where `MLoss` and `GLoss` start to produce `nan` values, should we stop there for that table?
```
Step 9000/25000 MLoss: 2.7555 GLoss: 0.9242 Sum: 3.6797
Step 9500/25000 MLoss: nan GLoss: nan Sum: nan
...
Step 24500/25000 MLoss: nan GLoss: nan Sum: nan
Step 25000/25000 MLoss: nan GLoss: nan Sum: nan
```

sdgym/synthesizers/clavaddpm.py
sarahmish: `None` is on purpose as in ClavaDDPM original code

sarahmish:
> During fit there is a point where `MLoss` and `GLoss` start to produce `nan` values, should we stop there for that table?

I lean towards keeping it as faithful to the original implementation as possible. Sometimes `nan` indicates that this model doesn't work for a particular dataset on this synthesizer.

pvk-developer (approved):

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1051 @@
...
+            if name in self._synthetic_pk:
+                df[self._primary_key[name]] = np.arange(len(df))
+
+            date_info = {}
+            for col, datetime_format in self._datetime_cols[name].items():
+                if datetime_format is None:
+                    datetime_array = df[df.notna()].astype(str).to_numpy()
+                    datetime_format = _guess_datetime_format_for_array(datetime_array)
+
+                days_since, earliest = calculate_days_since_earliest_date(df[col], datetime_format)
```
amontanez24, line 1081: Is this how they encoded datetimes in the paper, or is this our own encoding that we decided to go with?

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1051 @@
...
+import sys
+from collections import defaultdict
+from datetime import datetime, timedelta
+
+import numpy as np
+import pandas as pd
+import sklearn
+import torch
+from packaging.version import Version
+from pandas.core.tools.datetimes import _guess_datetime_format_for_array
```
amontanez24: is it safe to use this function since it is internal? Newer versions of pandas might remove it or change it significantly

sdgym/synthesizers/clavaddpm.py
sarahmish: It's not mentioned in the paper, but in their repo they have the [function](https://github.com/weipang142857/ClavaDDPM/blob/fa6ef4d7c6d2c584b45a398c4471bb51e0bb5e17/preprocess_utils.py#L10) defined. I assumed it's how they preprocess their data. In their repo, they also provided a few datasets before and after preprocessing

Here is an example from the berka dataset "card" table

**Before**
| card_id | disp_id | type | issued |
| ------- | -------- | ---- | ------- |
| 1005     | 9285    | classic |	931107 00:00:00 |
| 104       | 588      | classic |	940119 00:00:00 |
| 747       | 4915    | classic |	940205 00:00:00 |
| 70         | 439      | classic |	940208 00:00:00 |
| 577       | 3687    | classic | 	940215 00:00:00 |
| 377       | 2429    | classic |	940303 00:00:00 |

**After**
| card_id | disp_id | type | issued |
| ------- | -------- | ---- | ------- |
| 1005     | 9285    | 0 | 0 |
| 104       | 588      | 0 | 73 |
| 747       | 4915    | 0 | 90 |
| 70         | 439      | 0 | 93 |
| 577       | 3687    | 0 | 100 |
| 377       | 2429    | 0 | 116 |

sdgym/synthesizers/clavaddpm.py
sarahmish: I updated the function to our own guesser [here](https://github.com/sdv-dev/SDGym/pull/630/changes/8224ad184007ef0ae93a37c344b12cd3dc612e6c..72ecee05f6de65a82ad7dd07b84bf899f38a01d8)

sdgym/synthesizers/clavaddpm.py
```diff
@@ -31,6 +30,41 @@
...
+        return None
+
+    sample_size = min(sample_size, len(series))
+    sample = series.sample(sample_size, random_state=0)
+    sample = sample.astype(str).str.strip()
+    sample = sample[sample != '']
+
+    guesses = sample.apply(guess_datetime_format, dayfirst=dayfirst).dropna()
+    counts = Counter(guesses)
+    return top[0][0] if (top := counts.most_common(1)) else None
```
pvk-developer: If there are multiple datatime formats this should raise an error, mainly because the `calculate_days_since_earliest_date` will fail trying to cast those values:
```python
df = pd.DataFrame({
    'datetime_column': [
        '2025-01-01', 
        '2025-12-01', 
        '2025-12-01 22:25:30', 
        '2025-12-01 22:25:30', 
        '2025-12-01 22:25:30'
    ]
})

datetime_format = guess_array_datetime_format(df['datetime_column'])
print(datetime_format)
calculate_days_since_earliest_date(df['datetime_column'], datetime_format)
>>>
%Y-%m-%d %H:%M:%S
ValueError: time data '2025-01-01' does not match format '%Y-%m-%d %H:%M:%S'
```
Same happens if we do with lower precision:
```python
df = pd.DataFrame({
    'datetime_column': [
        '2025-01-01', 
        '2025-12-01', 
        '2025-12-01 22:25:30'
    ]
})

datetime_format = guess_array_datetime_format(df['datetime_column'])
print(datetime_format)
calculate_days_since_earliest_date(df['datetime_column'], datetime_format)
>>>
%Y-%m-%d
ValueError: unconverted data remains:  22:25:30
```

amontanez24 (approved):

pvk-developer (approved):

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1100 @@
...
+
+        return df[state['output_cols']].reset_index(drop=True)
+
+
+class ClavaDDPMSynthesizer(MultiTableBaselineSynthesizer):
+    """Custom wrapper for the ClavaDDPM synthesizer to make it work with SDGym."""
+
+    LOGGER = logging.getLogger(__name__)
+    _MODEL_KWARGS = None
+    _MODALITY_FLAG = 'multi_table'
```
R-Palazzo: You could remove this line since this is already set in the `MultiTableBaselineSynthesizer`:
https://github.com/sdv-dev/SDGym/blob/410fef0fc20c508d0b7d8f11b4696e9477ebeb19/sdgym/synthesizers/base.py#L142

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1201 @@
...
+
+    new_col_entry = {'type': 'discrete', 'size': len(set(parent_data_clusters.flatten()))}
+    parent_domain_dict[relation_cluster_name] = new_col_entry.copy()
+    child_domain_dict[relation_cluster_name] = new_col_entry.copy()
+
+    return parent_df_with_cluster, child_df_with_cluster, group_lengths_prob_dicts
+
+
+def match_tables(A, B, n_clusters=25, unique_matching=True, batch_size=100):
+    """Nearest-neighbour match of every row of ``A`` to a row of ``B``.
```
R-Palazzo: We could add from `pipeline_utils.py`

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1201 @@
...
+    child,
+    parents,
+    synthetic_tables,
+    id_cols,
+    n_clusters=1,
+    unique_matching=True,
+    batch_size=100,
+    no_matching=False,
+):
+    """Reconcile a child generated once per parent into a single table."""
```
R-Palazzo: Maybe also add from `pipeline_utils.py`

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1201 @@
...
+
+        self._parse_metadata(metadata)
+        self._metadata = metadata
+        self._fitted = False
+
+    def _parse_metadata(self, metadata):
+        meta = metadata if isinstance(metadata, dict) else metadata.to_dict()
+
+        self._table_names = list(meta['tables'])
+        self._primary_key = {}
```
R-Palazzo: Also define in l804

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1201 @@
...
+        self.diffusion_iterations = diffusion_iterations
+        self.batch_size = batch_size
+        self.lr = lr
+        self.weight_decay = weight_decay
+        self.sampling_batch_size = sampling_batch_size
+        self.num_matching_clusters = num_matching_clusters
+        self.matching_batch_size = matching_batch_size
+        self.unique_matching = unique_matching
+        self.no_matching = no_matching
+        self.max_categories = max_categories
```
R-Palazzo: This seems to never be used, should we get rid of it?

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1201 @@
...
+    matched_indices = []
+    matched_distances = []
+    for row_indices, row_distances in zip(indices, distances):
+        chosen, chosen_distance = None, 0.0
+        for candidate, distance in zip(row_indices, row_distances):
+            if int(candidate) not in used:
+                chosen, chosen_distance = int(candidate), float(distance)
+                break
+        if chosen is None:
+            chosen = next(j for j in range(len(B)) if j not in used)
```
R-Palazzo: Should we also update `chosen_distance` here?

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1201 @@
...
+    return anchor[0]
+
+
+@torch.no_grad()
+def _sample_step(diffusion, y):
+    """One reverse-diffusion pass conditioned on the label tensor ``y``.
+
+    Mirrors ``GaussianMultinomialDiffusion._sample`` but takes explicit
+    per-row labels instead of drawing them from the empirical distribution.
+    Similar to the original classifier-guided ``conditional_sample`` implemtation.
```
R-Palazzo:
```suggestion
    Similar to the original classifier-guided ``conditional_sample`` implementation.
```

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1201 @@
...
+            domain[col] = {'size': len(df[col].unique()), 'type': 'continuous'}
+    return domain
+
+
+def decode_id_values(values, encoder, column):
+    """Inverse-transform label-encoded id codes back to the original id values.
+
+    Sampling can create more rows than the original table, so codes past the
+    encoder's fitted range get fresh values: numeric ids continue after the
+    largest original id, other ids become new ``'{column}_{code}'`` strings.
```
R-Palazzo: Do we need to handle the numerical case here?

In l875 and l883 we're skipping the encoder for numerical ids

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1201 @@
...
+
+    for col_index, col in enumerate(original_parent_cols):
+        if col in parent_domain_dict:
+            if parent_domain_dict[col]['type'] == 'discrete':
+                parent_cat_cols.append((col_index, col))
+            else:
+                parent_num_cols.append((col_index, col))
+
+    parent_primary_key_index = original_parent_cols.index(parent_primary_key)
+    foreing_key_index = original_child_cols.index(foreign_key)
```
R-Palazzo, line 453:
```suggestion
    foreign_key_index = original_child_cols.index(foreign_key)
```

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1201 @@
...
+
+            column = cat_converted[:, col].reshape(-1, 1)
+            encoded_column = encoder.fit_transform(column)
+            cat_one_hot = np.concatenate((cat_one_hot, encoded_column), axis=1)
+
+        cat_one_hot[:, joint_cat_matrix_p_index:] = (
+            parent_scale * cat_one_hot[:, joint_cat_matrix_p_index:]
+        )
+
+    # Perform quantile normalization using QuantileTransformer
```
R-Palazzo: We removed `num_quantile` because it was not used?
https://github.com/weipang142857/ClavaDDPM/blob/fa6ef4d7c6d2c584b45a398c4471bb51e0bb5e17/pipeline_modules.py#L243

Maybe we should update the comment

sdgym/synthesizers/clavaddpm.py
```diff
@@ -0,0 +1,1201 @@
...
+    if joint_cat_matrix.shape[1] > 0:
+        cluster_data = np.concatenate((num_min_max, cat_one_hot, key_scaled), axis=1)
+    else:
+        cluster_data = np.concatenate((num_min_max, key_scaled), axis=1)
+
+    child_group_lengths = np.array([len(group) for group in child_group_data], dtype=int)
+    num_clusters = min(num_clusters, len(cluster_data))
+
+    init_param = 'k-means++'
+    if SKLEARN_VERSION.release[:2] < (1, 1):
```
R-Palazzo, line 544: `n_init='auto'` is also impacted by the sklearn version. It was introduced in scikit-learn 1.2:

https://scikit-learn.org/stable/whats_new/v1.2.html

At some point maybe we could just bump the minimum version for sklearn

sdgym/synthesizers/clavaddpm.py
sarahmish: This is additional logic from our side. ClavaDDPM requires id columns to be numeric values, so I added logic to encode non-numerical ids, which is why I added this part during decoding.

If the column is already numeric, the code defaults to expected ClavaDDPM behavior.

sdgym/synthesizers/clavaddpm.py
sarahmish: keeping the typo since it reflects their implementation

sarahmish:
Results for the financial data is [here](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?region=us-east-1&prefix=Debug/Sarah_ClavaDDPM_GCP/multi_table/SDGym_results_08_19_2026/&showversions=false).

Results for all other datasets can be found [here](https://us-east-1.console.aws.amazon.com/s3/buckets/sdgym-benchmark?region=us-east-1&prefix=Debug/ClavaDDPM_GCP_Sarah_all_datasets/multi_table/SDGym_results_08_20_2026/&showversions=false).

R-Palazzo (approved):
LGTM!

Looking at the GCP results, there are various errors for the private datasets (rel-bench datasets, instacart_marketbasket_ml, etc.).

To know whether the errors are coming from their code or from the extra logic we added, we would have to look into the log files, right? (Not the most convenient here)

